PR #54, opened 2024-02-12
misc: add env var to control log level

rpurdel:


skynet/logs.py
```diff
@@ -1,8 +1,8 @@
 import logging
 import sys
 from logging import Filter, LogRecord
-
 from uvicorn.logging import DefaultFormatter
+from env import log_level
```
saghul, line 5:
```suggestion
from .env import log_level
```

skynet/env.py
```diff
@@ -3,7 +3,8 @@
 
 is_mac = sys.platform == 'darwin'
 
-# modules
+# general
+log_level = os.environ.get('LOG_LEVEL', 'INFO').strip().upper()
```
saghul, line 7: Don't we want to default to debug?

skynet/logs.py
rpurdel: Used abs import to avoid any confusion

skynet/env.py
rpurdel: defaulted to debug

saghul (approved):

PR #56, opened 2024-02-13
feat: makefile

rpurdel:


Makefile
```diff
@@ -0,0 +1,35 @@
+ifneq (,$(wildcard ./.env))
+    include .env
+    export
```
saghul, line 3: Are any of the variables used in a subshell? If not I think you can drop the export.

Makefile
```diff
@@ -0,0 +1,35 @@
...
+endif
+
+GIT_HASH ?= $(shell git rev-parse --short HEAD)
+PLATFORMS ?= linux/amd64
+CACHE_DIR ?= /tmp/docker-cache
+
+login:
+	${DOCKER_LOGIN_CMD}
```
saghul, line 11: Don't we need to run this in a shell?

saghul (changes requested):

saghul (approved):

PR #59, opened 2024-02-16
summaries: refactor to use openai api model for inference

saghul:


skynet/modules/ttt/summaries/processor.py
```diff
@@ -45,4 +39,4 @@ async def process(job: Job) -> str:
 
     chain = load_summarize_chain(llm, chain_type="map_reduce", combine_prompt=prompt)
 
-    return await loop.run_in_executor(executor, chain.run, docs)
```
saghul, line 48: Nice!

skynet/env.py
```diff
@@ -6,14 +6,23 @@
...
 modules = supported_modules.intersection(enabled_modules)
 
 # models
 llama_path = os.environ.get('LLAMA_PATH')
-llama_n_gpu_layers = int(os.environ.get('LLAMA_N_GPU_LAYERS', 1 if is_mac else 40))
+llama_n_gpu_layers = int(os.environ.get('LLAMA_N_GPU_LAYERS', -1 if is_mac else 40))
 llama_n_batch = int(os.environ.get('LLAMA_N_BATCH', 512))
 
+# openai api
+openai_api_key = os.environ.get('OPENAI_API_KEY')
```
saghul, line 20: This is a good first step, but going forward we are going to need to figure this out differently. Likely by querying some vault service with some ID taken from the token.

skynet/modules/ttt/openai_api/app.py
```diff
@@ -5,7 +5,15 @@
 from skynet.env import llama_n_batch, llama_n_gpu_layers, llama_path
 from skynet.utils import dependencies, responses
 
-create_llama_cpp_app(Settings(model=llama_path, n_gpu_layers=llama_n_gpu_layers, n_batch=llama_n_batch, n_ctx=4096))
+create_llama_cpp_app(
+    Settings(
+        chat_format='llama-2',  # https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama_chat_format.py#L622C24-L622C31
+        model=llama_path,
+        n_batch=llama_n_batch,
+        n_ctx=4096,
```
saghul, line 13: Can we pl make this configurable?

skynet/modules/ttt/summaries/processor.py
```diff
@@ -1,42 +1,50 @@
...
-        model_path=llama_path,
-        temperature=0.01,
-        max_tokens=4096,
-        n_ctx=4096,
-        n_gpu_layers=llama_n_gpu_layers,
-        n_batch=llama_n_batch,
+    llm = ChatOpenAI(
+        api_key=openai_api_key
+        or 'placeholder',  # if no key is set, use a placeholder value to bypass validation, and allow the custom base url to be used
+        base_url=None if openai_api_key else openai_api_base_url,
```
saghul, line 25: Don't we need to pass headers here somehow? The OpenAI API app will also need the JWT.

quitrk: the openai app already has them

saghul: I mean, if we get a summary API request, with a token, and we then need to contact the openai compat app how are we passing the token?

quitrk: Right!

quitrk: Headers can be set on the langchain llm object only when initialising it, doing it at a later time has no effect.

Thus, there are two ways of making this work that I can think of, given the current setup of our auth flow:

1. Initialise the llm object every time a request is processed and have the JWT be forwarded from the client facing request all the way through. This shouldn't be expensive since we're not loading the model anymore here.

2. Have a separate auth flow between summaries api and the our openai api services, and thus add a second type of auth validation on the openai api. 

There's also a third choice, where we get rid of JWTs and introduce service to service auth, if we'll go down that path we still need to figure out how and if we should forward the auth details between the summaries and our openai api services

saghul: Perhaps we can fake it by using a class that implements "__str__" and can return different values per request.

It would only work if the headers are serialized late, but worth a shot IMHO.

saghul: _ _ str _ _ (no spaces but GH will format it otherwise)

quitrk: there's no need for that, if we just wanna pass through the jwt, we'll just init a new llm object on each processing flow, it has no performance impact

saghul: Ack. Let's go in this direction, since we want the app to be hit directly too, so we have to support tokens anyway.

quitrk: we need to discuss this a bit more, there are some other aspects to consider

skynet/modules/ttt/summaries/processor.py
```diff
@@ -1,48 +1,58 @@
...
-    llm = LlamaCpp(
-        model_path=llama_path,
-        temperature=0.01,
-        max_tokens=4096,
-        n_ctx=4096,
-        n_gpu_layers=llama_n_gpu_layers,
-        n_batch=llama_n_batch,
+    global llm
+
+    llm = ChatOpenAI(
```
saghul, line 22: This is the local app instance, right? Not the "real" OpenAI (or any other compatible service for that matter).

skynet/modules/ttt/summaries/processor.py
```diff
@@ -1,48 +1,58 @@
...
 
-    prompt = PromptTemplate(template=template, input_variables=["text"])
+    result = await chain.ainvoke(input={"input_documents": docs})
 
-    chain = load_summarize_chain(llm, chain_type="map_reduce", combine_prompt=prompt)
+    return result['output_text'].strip()
+
+
+async def process_open_ai(text: str, job_type: JobType, api_key: str) -> str:
+    llm = ChatOpenAI(
```
saghul, line 53: Would it make sense to have an LRU cache with there, rather than (potentially) build the same object over and over?

skynet/modules/ttt/summaries/prompts/action_items.py
```diff
@@ -1,6 +1,5 @@
-action_items_template = """
-<s>[INST] <<SYS>>
-Extract action items from the following text, if there are any.
-<</SYS>>
-{text} [/INST]
+action_items_system_message = """
+  The following text is a conversation between two or more people. If at the end of the conversation, there are future actions that need to be taken, please summarize them here. If there are no action items, please write "No action items."
```
saghul, line 2: I like this prompt better now.

skynet/modules/ttt/summaries/prompts/summary.py
```diff
@@ -1,6 +1 @@
-summary_template = """
-<s>[INST] <<SYS>>
-Provide a short summary of the following text.
-<</SYS>>
-{text} [/INST]
-"""
+summary_system_message = "Provide a concise summary of the following text."
```
saghul, line 1: Note for the future: let's add a "hint" parameter to the summaries API and have it be "conversation" or "text" and have different prompts.

skynet/modules/ttt/summaries/v1/router.py
```diff
@@ -21,12 +24,16 @@ async def get_action_items(payload: DocumentPayload) -> JobId:
...
 @api_version(1)
 @router.post("/summary")
-async def get_summary(payload: DocumentPayload) -> JobId:
+async def get_summary(payload: DocumentPayload, request: Request) -> JobId:
     """
     Starts a job to summarize the given payload.
     """
 
-    return await create_job(job_type=JobType.SUMMARY, payload=payload)
+    bearer_jwt: JWTBearer = await JWTBearer().__call__(request)
```
saghul, line 32: Where was auth happening up until now? Looks like the dependencies? Then aren't we auth-ing twice now?

skynet/modules/ttt/summaries/v1/router.py
```diff
@@ -21,12 +24,16 @@ async def get_action_items(payload: DocumentPayload) -> JobId:
...
+    decoded_jwt = await authorize(bearer_jwt)
+    customer_id = decoded_jwt.get("cid")
```
saghul, line 34: Let's catch potential problems and return a proper response.

skynet/modules/ttt/summaries/v1/router.py
```diff
@@ -21,12 +24,16 @@ async def get_action_items(payload: DocumentPayload) -> JobId:
...
+
+    return await create_job(job_type=JobType.SUMMARY, payload=payload, customer_id=customer_id)
```
saghul, line 36: `return await` is kinda an anti-pattern. If the function is async you can just return the result of calling the next function to save the creation of another future.

skynet/modules/ttt/summaries/v1/router.py
quitrk: that is done through the raise statements from `authorize`. isnt that enough?

skynet/modules/ttt/summaries/v1/router.py
quitrk: just decoding twice

skynet/modules/ttt/summaries/processor.py
```diff
@@ -1,48 +1,58 @@
...
+        api_key='placeholder',  # use a placeholder value to bypass validation, and allow the custom base url to be used
+        base_url=openai_api_base_url,
+        temperature=0,
     )
 
 
-async def process(job: Job) -> str:
-    text = job.payload.text
+async def process(text: str, job_type: JobType, model: ChatOpenAI = None) -> str:
```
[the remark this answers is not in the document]
quitrk: it's a wip....process_open_ai will be called once we query oci and get a valid open ai key

skynet/modules/ttt/summaries/processor.py
quitrk: yeah..it's just a langchain llm object

skynet/modules/ttt/summaries/jobs.py
```diff
@@ -58,12 +58,12 @@ async def restore_stale_jobs() -> list[Job]:
         await update_summary_queue_metric()
 
 
-async def create_job(job_type: JobType, payload: DocumentPayload) -> JobId:
+async def create_job(job_type: JobType, payload: DocumentPayload, customer_id: str) -> JobId:
```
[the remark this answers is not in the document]
quitrk: i didnt want to store the api key in redis

skynet/modules/ttt/openai_api/app.py
```diff
@@ -2,13 +2,21 @@
...
 
-from skynet.env import llama_n_batch, llama_n_gpu_layers, llama_path
-from skynet.utils import dependencies, responses
+from skynet.env import llama_n_batch, llama_n_ctx, llama_n_gpu_layers, llama_path
+from skynet.utils import openai_api_dependencies, responses
 
-create_llama_cpp_app(Settings(model=llama_path, n_gpu_layers=llama_n_gpu_layers, n_batch=llama_n_batch, n_ctx=4096))
+create_llama_cpp_app(
+    Settings(
+        chat_format='llama-2',  # https://github.com/abetlen/llama-cpp-python/blob/main/llama_cpp/llama_chat_format.py#L622C24-L622C31
```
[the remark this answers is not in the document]
quitrk: sure

skynet/modules/ttt/summaries/v1/router.py
quitrk: kind of, but more like decoding the jwt twice, since the pub key is cached anyway.

skynet/modules/ttt/summaries/prompts/summary.py
quitrk: I didn't want to change this significantly without testing it heavily first, since this is what's primarily being used out there

skynet/modules/ttt/summaries/processor.py
quitrk: i dont think this is costly enough to need that

skynet/modules/ttt/summaries/prompts/summary.py
saghul: We could make "text" the default and keep this prompt. Just an idea.

skynet/modules/ttt/summaries/v1/router.py
saghul: Hum, perhaps we can rework the dependency as a decorator to avoid the double parsing in the future.

skynet/modules/ttt/summaries/v1/router.py
saghul: That will return _some_ reply, but we can be more precise if we customize it, right? Like, the `cid` field is missing, etc.

skynet/auth/bearer.py
```diff
@@ -11,5 +11,6 @@ def __init__(self):
     async def __call__(self, request: Request):
         credentials: HTTPAuthorizationCredentials = await super().__call__(request)
 
-        if credentials and await authorize(credentials.credentials):
-            return credentials.credentials
+        request.state.decoded_jwt = await authorize(credentials.credentials)
```
saghul, line 14: You need to check if `credentials` exists, right? If not, auth will fail anyway.

skynet/main.py
```diff
@@ -44,8 +44,10 @@ async def lifespan(main_app: FastAPI):
         await summaries_startup()
 
     if 'summaries:executor' in modules:
+        from skynet.modules.ttt.openai_api.app import app as openai_api_app
         from skynet.modules.ttt.summaries.app import executor_startup as executor_startup
 
+        main_app.mount('/openai-api', openai_api_app)
```
saghul, line 50: We need to make sure the openai-api is not mounted twice.

skynet/modules/ttt/summaries/v1/models.py
```diff
@@ -31,10 +43,11 @@ class BaseJob(BaseModel):
 
 # since private fields are not serialized, use a different model with required internals
 class Job(BaseJob):
-    end: float | None = None
     created: float = Field(default_factory=time.time)
-    start: float | None = None
+    customer_id: str | None = None
```
saghul, line 47: Shouldn't we have the metadata instead?

skynet/modules/ttt/summaries/v1/router.py
```diff
@@ -1,32 +1,40 @@
...
 @api_version(1)
 @router.post("/action-items")
-async def get_action_items(payload: DocumentPayload) -> JobId:
+async def get_action_items(payload: DocumentPayload, request: Request) -> JobId:
     """
     Starts a job to extract action items from the given payload.
     """
 
-    return await create_job(job_type=JobType.ACTION_ITEMS, payload=payload)
+    customer_id = request.state.decoded_jwt.get("cid") if hasattr(request.state, 'decoded_jwt') else None
```
saghul, line 19: Isn't state a dictionary?

skynet/utils.py
```diff
@@ -4,13 +4,15 @@
...
 from skynet.auth.bearer import JWTBearer
-from skynet.env import bypass_auth
+from skynet.env import bypass_auth, bypass_openai_api_auth
 from skynet.logs import get_logger, uvicorn_log_config
 from skynet.modules.monitoring import FORCED_EXIT_COUNTER
 
 log = get_logger(__name__)
 
 dependencies = [] if bypass_auth else [Depends(JWTBearer())]
+openai_api_dependencies = [] if bypass_openai_api_auth else [Depends(JWTBearer())]
```
saghul, line 14: Isn't this redundant now? We should be able to use the same dependencies now. What'd be missing is the trick we discussed with having a UUID to skip auth.

skynet/auth/bearer.py
```diff
@@ -2,13 +2,17 @@
...
 from skynet.auth.jwt import authorize
+from skynet.env import app_uuid
 
 
 class JWTBearer(HTTPBearer):
     def __init__(self):
         super().__init__(auto_error=True)
 
     async def __call__(self, request: Request):
+        if request.headers.get('X-Skynet-UUID') == app_uuid:
```
saghul, line 13: 🔥

skynet/auth/openai.py
```diff
@@ -0,0 +1,29 @@
...
+        raise RuntimeError(f'Error loading credentials file: {e}')
+
+
+def setup_credentials():
+    if openai_credentials_file:
+        open_yaml(openai_credentials_file)
+        file_watcher = FileWatcher(openai_credentials_file, lambda: open_yaml(openai_credentials_file))
+
+        file_watcher.start()
+    else:
```
saghul, line 24: This should not be mandatory.

skynet/auth/openai.py
```diff
@@ -0,0 +1,29 @@
...
+    except Exception as e:
+        raise RuntimeError(f'Error loading credentials file: {e}')
+
+
+def setup_credentials():
+    if openai_credentials_file:
+        open_yaml(openai_credentials_file)
+        file_watcher = FileWatcher(openai_credentials_file, lambda: open_yaml(openai_credentials_file))
+
+        file_watcher.start()
```
saghul, line 23: When are we going to stop this?

skynet/auth/openai.py
```diff
@@ -0,0 +1,29 @@
...
+
+from skynet.env import openai_credentials_file
+from skynet.modules.file_watcher import FileWatcher
+
+credentials = dict()
+
+
+def open_yaml(file_path):
+    try:
+        with open(file_path, 'r') as stream:
```
saghul, line 11: This operation is blocking, we should look into using aio-files for file i/o.

skynet/auth/openai.py
```diff
@@ -0,0 +1,29 @@
...
+        open_yaml(openai_credentials_file)
+        file_watcher = FileWatcher(openai_credentials_file, lambda: open_yaml(openai_credentials_file))
+
+        file_watcher.start()
+    else:
+        raise RuntimeError('The OpenAI credentials file must be set')
+
+
+def get_credentials(customer_id):
+    return credentials.get(customer_id)
```
saghul, line 29: Let's return a canary value instead of an exception? Like a mock object with None key and model.

skynet/modules/file_watcher.py
```diff
@@ -0,0 +1,40 @@
...
+log = get_logger(__name__)
+
+
+class FileWatcher:
+    def __init__(self, file_path, callback=None, refresh_delay_secs=file_refresh_interval):
+        self.background_task = None
+        self.file_path = file_path
+        self.callback = callback
+        self.refresh_delay_secs = refresh_delay_secs
+        self._previous_modified_time = os.stat(file_path).st_mtime
```
saghul, line 17: This is a blocking operation. It can also throw an exception.

skynet/modules/file_watcher.py
```diff
@@ -0,0 +1,40 @@
...
+    async def poll_for_changes(self):
+        while True:
+            try:
+                await asyncio.sleep(self.refresh_delay_secs)
+                self.watch()
+            except Exception as e:
+                log.error(f'Error while polling for file changes: {e}')
+                break
+
+    def start(self):
```
saghul, line 39: I guess there should be a stop here...

skynet/modules/ttt/summaries/app.py
```diff
@@ -25,13 +27,17 @@ async def app_startup():
 
 
 async def executor_startup():
+    setup_credentials()
+    log.info('Credentials set. Watching for changes...')
```
saghul, line 31: I think we should log this from the credentials thing, since it might not actually be used.

skynet/modules/ttt/summaries/jobs.py
```diff
@@ -119,7 +121,17 @@ async def run_job(job: Job) -> None:
         exit_task = asyncio.create_task(exit_on_timeout())
 
         try:
-            result = await process(job)
+            if job.metadata.customer_id:
+                options = get_credentials(job.metadata.customer_id)
+
+                if options and options.get('api_key'):
```
saghul, line 127:
```suggestion
                if options and options.get('api_key', None) is not None:
```

skynet/modules/ttt/summaries/jobs.py
```diff
@@ -119,7 +121,17 @@ async def run_job(job: Job) -> None:
...
+                    log.info(f"Forwarding inference to OpenAI for customer {job.metadata.customer_id}")
+
+                    result = await process_open_ai(
+                        job.payload, job.type, options['api_key'], model_name=options.get('model_name')
```
saghul, line 131:
```suggestion
                        job.payload, job.type, options['api_key'], model_name=options.get('model_name', None)
```

skynet/modules/ttt/summaries/processor.py
```diff
@@ -1,48 +1,73 @@
...
-    executor = ThreadPoolExecutor(max_workers=1)
-
-    llm = LlamaCpp(
-        model_path=llama_path,
-        temperature=0.01,
-        max_tokens=4096,
-        n_ctx=4096,
-        n_gpu_layers=llama_n_gpu_layers,
-        n_batch=llama_n_batch,
+    global llm
```
saghul, line 32: Is this initialize necessary? Can't we just create it at imprt time?

saghul (commented):
Left some comments, PTAL!

skynet/auth/openai.py
quitrk: when the process stops. Is there other time when we'd want it to stop

skynet/auth/openai.py
quitrk: but it's not returning an exception, that's why I used `.get`

skynet/modules/ttt/summaries/processor.py
quitrk: if we move it like that, we need to also better isolate the summaries:dispatcher module to make sure it's not importing it indirectly.

skynet/auth/openai.py
saghul: Not really, in general I prefer explicit vs implicit. No big deal though.

skynet/modules/ttt/summaries/processor.py
saghul: Gotcha. No strong opinion then.

saghul (commented):
LGTM, excellent work! Since I opened the PR you need to approve it yourself lol

saghul:
Pl squash-merge.

PR #61, opened 2024-02-20
Get time from chunk

rpurdel:


skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -28,20 +28,19 @@ def __init__(
         # before starting to drop them
         add_max_silent_chunks: int = 1,
         final_after_x_silent_chunks: int = 2,
-        force_final_duration_threshold: int = 29,
+        force_final_duration_threshold: int = 15,
         perform_final_after_silent_seconds: float = 0.8,
     ):
-        self.transcription_id = str(uuid6.uuid7())
```
saghul, line 34: These are equivalent, except that the previous version is better because it's v7 instead of v1.

skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -57,65 +56,86 @@ def _extract_transcriptions(
         final = ''
         interim = ''
         # return a final and flush the working audio if too many silent chunks
-        # or if the buffer is over the max length threshold
+        # or if the working audio is over the max duration threshold
         if (
             self.silent_chunks >= self.final_after_x_silent_chunks
             or utils.convert_bytes_to_seconds(self.working_audio) > self.force_final_duration_threshold
         ):
             if ts_result.text.strip() != '':
```
saghul, line 64: Better check for the length.

skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -57,65 +56,86 @@ def _extract_transcriptions(
...
-                results.append(self.get_response_payload(ts_result.text, True))
-                self.all_final += ts_result.text
-                self.reset(True)
+                start_timestamp = int(ts_result.words[0].start * 1000) + self.working_audio_starts_at
```
saghul, line 65: What does start contain?

skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -57,65 +56,86 @@ def _extract_transcriptions(
...
         if working_audio_duration >= 1 and not self.long_silence:
             return True
         return False
 
-    async def process(self, chunk: bytes) -> List[utils.TranscriptionResponse] | None:
+    async def process(self, chunk: bytes, chunk_timestamp: int) -> List[utils.TranscriptionResponse] | None:
         self.chunk_count += 1
-        chunk_duration = utils.convert_bytes_to_seconds(chunk)
-        self.user_time += chunk_duration
+        if self.chunk_duration == 0:
```
saghul, line 129: This reads weird because we are not taking the duration straight from the chunk itself.

skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -136,37 +156,44 @@ def add_to_store(self, chunk: bytes):
...
+        self, transcription: str, start_timestamp: int, final_audio: bytes | None = None, final: bool = False
+    ) -> utils.TranscriptionResponse:
         return utils.TranscriptionResponse(
-            id=self.transcription_id,
+            id=str(utils.uuid_from_time(start_timestamp)),
             participant_id=self.participant_id,
-            ts=self.transcription_ts if final else utils.now(),
+            ts=start_timestamp,
             text=transcription,
+            audio=base64.b64encode(final_audio).decode('ASCII') if final_audio else '',
```
saghul, line 185: Can we skip this if the log level is not debug, or maybe even another env variable? I can imagive privacy concers, but also performance.

skynet/modules/stt/streaming_whisper/app.py
```diff
@@ -24,7 +25,7 @@ async def websocket_endpoint(websocket: WebSocket, meeting_id: str, auth_token:
                 log.info(f'Received disconnect message for {meeting_id}')
                 ws_connection_manager.disconnect(meeting_id)
                 break
-            await ws_connection_manager.process(meeting_id, chunk)
```
saghul, line 27: Why don't we generate a UUID v7 right here right now? (rather in the process function I guess). Since it's the processing time. We use the time to then generate the ID... We could generate the ID and extract the time if we need to. But even generating the ID and then getting a new now() timestamp should be fine for our purpose.

skynet/modules/stt/streaming_whisper/state.py
```diff
@@ -28,20 +28,19 @@ def __init__(
...
         self.participant_id = participant_id
         self.silent_chunks = 0
+        self.chunk_count = 0
         self.working_audio = b''
         self.lang = lang
-        self.all_final = ''
         self.long_silence = False
-        self.chunk_count = 0
-        self.user_time = 0.0
+        self.chunk_duration = 0.0
```
saghul, line 43: Seeing a number of chunk related bookeeping we have, would it make sense to create a little `Chunk` class wrapping the binary data, with some helper methods perhaps?

saghul (commented):
Left some comments, but you'll likely have to walk me though this...

skynet/modules/stt/streaming_whisper/utils/utils.py
```diff
@@ -286,3 +285,27 @@ def get_lang(lang: str, short=True) -> str:
         split_key = lang.split('-')[0]
         return LANGUAGES.get(split_key, 'english').lower().strip()
     return lang.lower().strip()
+
+
+def uuid_from_time(time_arg: int) -> uuid.UUID:
```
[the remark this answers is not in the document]
rpurdel: replaced it with uuid7

skynet/modules/stt/streaming_whisper/state.py
rpurdel: added a chunk class

skynet/modules/stt/streaming_whisper/state.py
rpurdel: the floating seconds of when the word starts since the beginning of the working audio buffer

skynet/modules/stt/streaming_whisper/state.py
rpurdel: I already defined an env var which is off by default

pyproject.toml
```diff
@@ -22,7 +22,7 @@ async-lru = "2.0.4"
 boto3 = "^1.28.56"
 fastapi = "0.109"
 fastapi-versionizer = "3.0.4"
-faster-whisper = "0.10.0"
+faster-whisper = "v0.10.1"
```
saghul, line 25: I already updated this in another pr, can you pl rebase?

saghul (approved):

saghul:
Great stuff @rpurdel ! You'll need to rebase so resolve some conflicts from the faster0-whisper package update, LGTM otherwise!

PR #65, opened 2024-03-07
Streaming whisper flush worker

rpurdel:


skynet/modules/stt/streaming_whisper/connection_manager.py
```diff
@@ -14,9 +15,14 @@
 
 class ConnectionManager:
     connections: dict[str, MeetingConnection]
+    flush_after_ms: int
 
     def __init__(self):
         self.connections: dict[str, MeetingConnection] = {}
+        self.flush_after_ms = 2000
+        log.info('Starting the flush audio worker')
```
saghul, line 23: Is this log really useful?

skynet/modules/stt/streaming_whisper/connection_manager.py
```diff
@@ -14,9 +15,14 @@
 
 class ConnectionManager:
     connections: dict[str, MeetingConnection]
+    flush_after_ms: int
 
     def __init__(self):
         self.connections: dict[str, MeetingConnection] = {}
+        self.flush_after_ms = 2000
```
saghul, line 22: Make this a constant in all caps outside of the class please, since we are not changing the value withint the class anyway.

skynet/modules/stt/streaming_whisper/connection_manager.py
```diff
@@ -14,9 +15,14 @@
...
+        log.info('Starting the flush audio worker')
+        loop = asyncio.get_running_loop()
+        loop.create_task(self.flush_working_audio_worker())
```
saghul, line 25: I think it would be best to create the task lazily in the `connect` method, if it didn't already exist.

saghul (commented):
Left some comments, PTAL!

saghul (approved):

PR #74, opened 2024-04-30
feat: allow docker tag to be specified as parameter in Makefile

aaronkvanmeerten:


saghul (approved):

saghul:
Do we still want this? If so, it needs a rebase / do over.

PR #77, opened 2024-05-14
deps: update llama-cpp-python

saghul:


skynet/env.py
```diff
@@ -6,6 +6,17 @@
 
 is_mac = sys.platform == 'darwin'
 
+
+# utilities
+def tobool(val: str | None):
```
quitrk, line 11: you might wanna consider moving this to https://github.com/jitsi/skynet/blob/master/skynet/utils.py

quitrk (approved):

aaronkvanmeerten (approved):

saghul:
@quitrk I merged as-is because moving tobool to utils would mean env.py imports utils.py, which in turn imports env.py leading to import errors because the module is not fully initialized.

We can probably reorganize that a bit, but IMHO not worth it for a single function. If things get more hairy we can of course separate things more semantically. For example, app_uuid doesn't really belong in env, but there is no other good place for it, maybe utils? I hate utils modules, them being a mixed bag of everything...

PR #80, opened 2024-05-16
feat: allow websocket header auth from non-js clients

rpurdel:


skynet/modules/stt/streaming_whisper/utils/utils.py
```diff
@@ -305,3 +305,12 @@ def get(self, time_arg_millis: int = None) -> UUID:
         uuid_int = (timestamp_ms & 0xFFFFFFFFFFFF) << 80
         uuid_int |= secrets.randbits(76)
         return UUID(int=uuid_int, version=7)
+
+
+def get_jwt(ws_headers, ws_url_param=None) -> str:
+    jwt = ws_url_param if ws_url_param else ''
```
saghul, line 311: Shouldn't we return early if we find it in the URL?

skynet/modules/stt/streaming_whisper/connection_manager.py
```diff
@@ -25,7 +25,10 @@ def __init__(self):
 
     async def connect(self, websocket: WebSocket, meeting_id: str, auth_token: str | None):
         if not bypass_auth:
-            authorized = await authorize(auth_token)
+            jwt_token = utils.get_jwt(websocket.headers, auth_token)
+            authorized = await authorize(
```
saghul, line 29: Why this style with the 3 lines and dangling comma, when it fits in a si gel one?

skynet/modules/stt/streaming_whisper/utils/utils.py
rpurdel: not really, I would like it to always pick the auth header if present. The dict key check is very fast anyway

saghul: My comment is not about performance, but rather this looks weird in case you have both. If we prefer the header, let's check there first, else return the other. It makes the code more straightforward, no?

saghul (approved):
LGTM with a nit

saghul (approved):

PR #83, opened 2024-05-30
wip

quitrk:


Dockerfile
```diff
@@ -15,16 +15,27 @@ COPY docker/rootfs/ /
...
+    sh cmake.sh --skip-license --prefix=/usr/local && \
+    rm cmake.sh
+
+ENV LLAMA_CPP_RELEASE=b3033
+
+RUN \
+    git clone https://github.com/ggerganov/llama.cpp.git --depth=1 --branch $LLAMA_CPP_RELEASE && \
+    cd llama.cpp && \
+    cmake -B build -DCMAKE_BUILD_TYPE=Release -DLLAMA_CUDA=ON -DLLAMA_NATIVE=OFF && \
+    cmake --build build --config Release --target server -j`getconf _NPROCESSORS_ONLN`
```
saghul, line 32:
```suggestion
    cmake --build build --target server -j`getconf _NPROCESSORS_ONLN`
```

saghul: The `--config` option is for multi-config targets, but Unix Makefiles (the default for Unix systems) is not that.

Dockerfile
```diff
@@ -46,7 +57,7 @@ COPY --chown=jitsi:jitsi docker/run-skynet.sh /opt/
 RUN \
     apt-dpkg-wrap apt-key adv --keyserver keyserver.ubuntu.com --recv-keys F23C5A6CF475977595C89F51BA6932366A755776 && \
     apt-dpkg-wrap apt-get update && \
-    apt-dpkg-wrap apt-get install -y python3.11 python3.11-venv tini libgomp1 && \
+    apt-dpkg-wrap apt-get install -y python3.11 python3.11-venv tini libgomp1 libcurl4-openssl-dev && \
```
saghul, line 60: For prod you don't want the `-dev` library, just the runtime.

Dockerfile
```diff
@@ -15,16 +15,27 @@ COPY docker/rootfs/ /
 RUN \
     apt-dpkg-wrap apt-key adv --keyserver keyserver.ubuntu.com --recv-keys F23C5A6CF475977595C89F51BA6932366A755776 && \
     apt-dpkg-wrap apt-get update && \
-    apt-dpkg-wrap apt-get install -y build-essential python3.11 python3.11-venv && \
+    apt-dpkg-wrap apt-get install -y git wget build-essential git python3.11 python3.11-venv && \
```
saghul, line 18: git is there twice.

saghul: Also, don't we need the curl library, so the model tester works?

compose-dev.yaml
```diff
@@ -0,0 +1,14 @@
+services:
```
saghul, line 1: Nice!

skynet/env.py
```diff
@@ -19,22 +19,21 @@ def tobool(val: str | None):
...
-# Some formats are auto-detected: https://github.com/abetlen/llama-cpp-python/blob/c50d3300d2a09c98765be7f2c05b7e4fd0b4232e/llama_cpp/llama_chat_format.py#L724
-model_chat_format = os.environ.get('MODEL_CHAT_FORMAT')
 llama_path = os.environ.get('LLAMA_PATH')
 llama_n_ctx = int(os.environ.get('LLAMA_N_CTX', 8192))
 llama_n_gpu_layers = int(os.environ.get('LLAMA_N_GPU_LAYERS', -1 if is_mac else 40))
 llama_n_batch = int(os.environ.get('LLAMA_N_BATCH', 512))
 
 # openai api
-openai_api_base_url = os.environ.get('OPENAI_API_BASE_URL', 'http://localhost:8000/openai-api/v1')
+openai_api_server_dir = os.environ.get('OPENAI_API_SERVER_DIR', '/app/llama.cpp/server')
```
saghul, line 34: Let's take a path to the binary instead. Is there a reason why the directory is necessary?

skynet/env.py
```diff
@@ -19,22 +19,21 @@ def tobool(val: str | None):
...
-supported_modules = {'summaries:dispatcher', 'summaries:executor', 'openai-api', 'streaming_whisper'}
+supported_modules = {'summaries:dispatcher', 'summaries:executor', 'streaming_whisper'}
 enabled_modules = set(os.environ.get('ENABLED_MODULES', 'summaries:dispatcher,summaries:executor').split(','))
 modules = supported_modules.intersection(enabled_modules)
 file_refresh_interval = int(os.environ.get('FILE_REFRESH_INTERVAL', 30))
 
 # models
-
-# Some formats are auto-detected: https://github.com/abetlen/llama-cpp-python/blob/c50d3300d2a09c98765be7f2c05b7e4fd0b4232e/llama_cpp/llama_chat_format.py#L724
-model_chat_format = os.environ.get('MODEL_CHAT_FORMAT')
```
saghul, line 30: Hum, doesn't llama.cpp get anything like that?

Dockerfile
quitrk: that's a relic from some different attempt to make the server work, will remove

skynet/env.py
quitrk: forgot to rename it, it is a path to the binary

skynet/env.py
quitrk: it infers it from the model metadata if the info is supplied, which it is for our usecases

Dockerfile
quitrk: actually that was three times, there was another git installed above :).

Wdym about the model tester?

saghul: Don't we need libcurl so the demo page works?

skynet/env.py
saghul: Sounds good. If there is a way to set it, we might eventually need it in order to test mistral or whatever, but at this point llama-3 seems to be the way to go anyway.

skynet/env.py
```diff
@@ -31,7 +31,7 @@ def tobool(val: str | None):
 llama_n_batch = int(os.environ.get('LLAMA_N_BATCH', 512))
 
 # openai api
-openai_api_server_dir = os.environ.get('OPENAI_API_SERVER_DIR', '/app/llama.cpp/server')
+openai_api_server_path = os.environ.get('OPENAI_API_SERVER_PATH', '/app/llama.cpp/server')
```
saghul, line 34: I'd call lthis `llama_server_path` since it's very specific to llama.cpp

skynet/modules/ttt/summaries/jobs.py
```diff
@@ -71,7 +71,7 @@ async def create_job(job_type: JobType, payload: DocumentPayload, metadata: Docu
 
     await db.set(job_id, Job.model_dump_json(job))
 
-    log.info(f"Created job {job.id}.")
+    log.info(f"Created job {job}.")
```
saghul, line 74: Unrelated change? What does this log now?

quitrk: 'temp logging' commit, ill remove it soon

saghul (approved):

PR #87, opened 2024-06-13
feat: add support for Azure OpenAI

quitrk:


credentials.yaml.sample
```diff
@@ -1,4 +1,19 @@
...
+            secret: test_secret
+            type: AZURE_OPENAI
+        OPENAI:
+            customerId: testCustomerId
+            enabled: false
+            metadata:
+                model: gpt-3
+            secret: test_secret
+            type: OPENAI
+    customerId: testCustomerId
```
saghul, line 19: stray line?

skynet/auth/openai.py
```diff
@@ -40,4 +47,13 @@ async def setup_credentials():
 
 
 def get_credentials(customer_id):
-    return credentials.get(customer_id, {}) or {}
+    customer_credentials = credentials.get(customer_id, {}) or {}
+    multiple_credentials = customer_credentials.get('credentialsMap')
+
+    if multiple_credentials:
+        result = [val for val in multiple_credentials.values() if val['enabled'] == True]
```
saghul, line 54:
```suggestion
        result = [val for val in multiple_credentials.values() if val['enabled']]
```

saghul: No need to check explicitly for `True`, unless that thing can have any other truthy values?

saghul (approved):
2 minor comments, great work!

credentials.yaml.sample
quitrk: wdym? it's an actual field in the yaml

quitrk: though we don't really use it so we could get rid of it for the sample

PR #92, opened 2024-08-20
Upgrade faster-whisper

rpurdel:


Makefile
```diff
@@ -11,22 +11,18 @@ _login:
 
 build-summaries : _login
 	docker buildx build \
-	--build-arg="BASE_IMAGE_BUILD=nvidia/cuda:12.3.0-devel-ubuntu20.04" \
-	--build-arg="BASE_IMAGE_RUN=nvidia/cuda:12.3.0-runtime-ubuntu20.04" \
+	--build-arg="BASE_IMAGE_BUILD=nvidia/cuda:12.2.2-cudnn8-devel-ubuntu22.04" \
```
saghul, line 14: Shouldn't we drop these now that they match the defaults?

Makefile
```diff
@@ -11,22 +11,18 @@ _login:
...
+	--build-arg="BASE_IMAGE_RUN=nvidia/cuda:12.2.2-cudnn8-runtime-ubuntu22.04" \
 	--progress plain \
 	--push \
 	--platform ${PLATFORMS} \
-	--cache-from type=local,src=${CACHE_DIR} \
```
saghul, line 19: Why drop these?

docker/rootfs/etc/apt/sources.list.d/python-ppa.list
```diff
@@ -1,2 +1,2 @@
-deb https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu focal main
-deb-src https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu focal main
+deb https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy main
+deb-src https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy main
```
saghul, line 2: We don't need the src, now that I see it.

Makefile
rpurdel: not really working

saghul: 👍

saghul (approved):
🚀  No code changes necessary when updating faster-whisper? Impressive!

PR #97, opened 2024-08-28
feat: add integration support for vox implant

quitrk:


skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
+import base64
```
saghul, line 1: In my sample I used an alternativr which is much faster using SIMD instructions, maybe use that for some free perf?

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+from skynet.logs import get_logger
+from skynet.modules.stt.streaming_whisper.utils import utils
+from skynet.modules.stt.vox.connection_manager import ConnectionManager
+from skynet.modules.stt.vox.decoder import PcmaDecoder
+
+log = get_logger(__name__)
+
+ws_connection_manager = ConnectionManager()
+app = FastAPI()
+decoder = PcmaDecoder()
```
saghul, line 14: In general, it makes more sense to use a decoder per connection, since they might be using different sample rates / codecs and audio samples would interleave.

saghul: Even if that's not the case yet, let's make it a local variable in the websocket handler please.

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+    sampling_rate = 8000
+
+    while True:
+        try:
+            ws_data = await websocket.receive_json()
+
+            event = ws_data.get('event')
+
+            if event == 'start':
+                sampling_rate = ws_data['start']['mediaFormat']['sampleRate']
```
saghul, line 32: Try / catch and use some default?

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+
+            if event == 'media':
+                media = ws_data['media']
+                participant_id: str = media['tag']
+
+                if participant_id not in data_map:
+                    header = (participant_id.encode() + '|en'.encode()).ljust(60, b'\0')
```
saghul, line 39: I thought we were going to not do this anynore and handle metadata in JSON.

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+                    data_map[participant_id] = dict(raw=b'', chunks=0, header=header)
+
+                payload = base64.b64decode(media['payload'])
+                participant = data_map[participant_id]
+
+                participant['raw'] += payload
+                participant['chunks'] += 1
+
+                if participant['chunks'] == 50:  # 50 chunks = 1 second
```
saghul, line 48: Why are we buffering here? Does the process function require that the audio already be buffered?

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+                    frames = decoder.decode(participant['raw'], media['timestamp'])
+
+                    decoded_raw = b''
```
saghul, line 51: In the next 5 lines we are copying buffers several times, there likely is a better way? I think we can pre-allocate a buffer and then copy the data inside, then just resample inplace?

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,72 @@
...
+                    for frame in frames:
+                        decoded_raw += bytes(frame.planes[0])
+
+                    if sampling_rate != 16000:
+                        decoded_raw = utils.resample_audio(decoded_raw, sampling_rate, 16000)
+
+                    await ws_connection_manager.process(
```
saghul, line 58: Doesn't this introduce a backlog, since we are not processing chunks while we are processing them? Once we have a large enough sample it feels like we would trancribe that buffer and process incoming chunks concurrently?

skynet/modules/stt/vox/connection_manager.py
```diff
@@ -0,0 +1,43 @@
...
+
+from skynet.logs import get_logger
+from skynet.modules.stt.streaming_whisper.connection_manager import ConnectionManager as BaseConnectionManager
+from skynet.modules.stt.streaming_whisper.utils.utils import TranscriptionResponse
+
+log = get_logger(__name__)
+
+
+class ConnectionManager(BaseConnectionManager):
+    async def process(self, session_id: str, chunk: bytes, timestamp: int, tag: str):
```
saghul, line 13: This is called chunk but we don't call it per chunk, but with a buffer, so let's call it `buffer` or `data` to avoid confusion.

saghul (commented):
Great stuff! Left some comments, PTAL!

As to the app name / intended purpose, this will likely become the v2 of streaming whisper, right?

We just use the Vox payload because we like it.

To that end, can you add some README on the feature folder with the payload definition and intended use?

requirements.txt
```diff
@@ -7,8 +7,8 @@ async-lru==2.0.4 ; python_version >= "3.11" and python_version < "3.12"
 async-timeout==4.0.3 ; python_version >= "3.11" and python_full_version <= "3.11.2"
 attrs==24.2.0 ; python_version >= "3.11" and python_version < "3.12"
 av==12.3.0 ; python_version >= "3.11" and python_version < "3.12"
-boto3==1.34.156 ; python_version >= "3.11" and python_version < "3.12"
-botocore==1.34.156 ; python_version >= "3.11" and python_version < "3.12"
+boto3==1.35.7 ; python_version >= "3.11" and python_version < "3.12"
```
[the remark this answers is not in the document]
quitrk: this was due to a `poetry update` request

skynet/modules/stt/vox/app.py
quitrk: well this is just compatibility with the streaming whisper logic, we're gonna have to decide how we want to move away from that..

skynet/modules/stt/vox/app.py
quitrk: the process function also does silence detection through silero for each chunk it receives sending 20ms will yield incorrect results. If we want to move away from how streaming whisper does things, we'l have to rethink this

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,75 @@
...
+    while True:
+        try:
+            ws_data = await websocket.receive_json()
+
+            event = ws_data.get('event')
+
+            if event == 'start':
+                try:
+                    sampling_rate = ws_data['start']['mediaFormat']['sampleRate']
+                except:
```
saghul, line 34: Handle KeyError directly pl

skynet/modules/stt/vox/app.py
```diff
@@ -0,0 +1,75 @@
...
+                participant['chunks'] += 1
+
+                if participant['chunks'] == 50:  # 50 chunks = 1 second
+                    frames = decoder.decode(participant['raw'], media['timestamp'])
+
+                    decoded_raw = b''
+                    for frame in frames:
+                        decoded_raw += bytes(frame.planes[0])
+
+                    if sampling_rate != 16000:
```
saghul, line 58: Let's put the 16k in a constant and document that it's because whisper needs it that way pl

saghul (approved):
LGTM! Can you please undo the llama.cpp update, since it's in another commit?

quitrk:
> LGTM! Can you please undo the llama.cpp update, since it's in another commit?

sure, I still need to refactor the resampling. Also, this won't be merged until we figure out the next steps

saghul (approved):
Great work!

PR #99, opened 2024-09-10
deps: update llama.cpp

quitrk:
- this fixes timeouts when running llama 3.1
- bypass storing failed jobs in redis if they were ran with external processor

### DIFF
--- a/skynet/modules/ttt/summaries/jobs.py
+++ b/skynet/modules/ttt/summaries/jobs.py
@@ -137,6 +137,9 @@ async def _run_job(job: Job) -> None:
 
     await update_job(job_id=job.id, start=start, status=JobStatus.RUNNING, worker_id=worker_id)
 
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Out of curiosity what commit in llama.cpp fixed it?

quitrk:
> Out of curiosity what commit in llama.cpp fixed it?

I traced the fix back to two commits.

https://github.com/ggerganov/llama.cpp/commit/4b9afbbe9037f8a2d659097c0c7d9fce32c6494c
https://github.com/ggerganov/llama.cpp/commit/06658ad7c37f440502de2b9486ce43c47b4ec710

Now, the second one seems mac related, so I'll have to test some more on the A10 after I deploy in stage

saghul:
🙌🙌

PR #100, opened 2024-09-18
fix: wait for openai api server to be up before running jobs

quitrk:


skynet/modules/ttt/openai_api/app.py
```diff
@@ -38,6 +39,19 @@ def initialize():
         log.info(f'OpenAI API server started from {openai_api_server_path}')
 
 
+async def is_started():
+    # https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md#get-health-returns-heath-check-result
+    try:
+        response = await http_client.get(f'http://localhost:{openai_api_server_port}/health')
```
saghul, line 45: Don't we have the base URL in an env variable?

saghul (approved):

skynet/modules/ttt/openai_api/app.py
quitrk: for some reason i looked at that declaration and ignored its existence

PR #101, opened 2024-09-18
fix: resolve issues with jobs failing with llama 3.1

quitrk:
Most of the issues with inference failures solved after upgrading llama.cpp and updating the prompting to correctly specify the system role 

Added a new script for running skynet jobs. This will be incorporated in some automation flow 
Added a new hint type for emails
Updated llama.cpp and langchain & co to the latest

There are still cases of timeouts, but these are due to hallucinations of the llm which go on until the context window is full. Since llama 3.0 had a lower context window, the rest server would simply return whatever was generated up until the point the 8192 tokens were filled, while for llama 3.1, since we're using a much larger window, our timeout intervenes before the context window gets to be filled.


pyproject.toml
```diff
@@ -35,9 +35,11 @@ torchaudio = ">=2.0.1,<2.1.0"
 uvicorn = {extras = ["standard"], version = "0.29.0"}
 uuid6 = "^2024.1.12"
 openai = "^1.12.0"
-langchain-openai = "^0.0.6"
 pyyaml = "^6.0.1"
 aiofiles = "^23.2.1"
+pydantic = "^2.9.1"
+langchain = "^0.3.0"
+langchain-openai = "^0.2.0"
```
saghul, line 42: Should we update openai past 1.12 ?

run-skynet-requests.py
```diff
@@ -0,0 +1,89 @@
...
+
+max_requests = int(args.max_requests)
+base_url = args.url
+sleep_time = int(args.sleep)
+
+
+async def main():
+    session = aiohttp.ClientSession()
+
+    async def post(job_type, data):
```
saghul, line 35: Does it make sense to add jwt support to test it against stage for example?

saghul (approved):

run-skynet-requests.py
quitrk: it does, i actually wanted to do that but kinda forgot, ty for reminding me

pyproject.toml
quitrk: name = "langchain-openai"
version = "0.2.0"
description = "An integration package connecting OpenAI and LangChain"
optional = false
python-versions = "<4.0,>=3.9"
files = [
    {file = "langchain_openai-0.2.0-py3-none-any.whl", hash = "sha256:9a1a69ba0706f23ec2941096ead0bc39202cac0e9782a5d6c8d92cb2280c2759"},
    {file = "langchain_openai-0.2.0.tar.gz", hash = "sha256:441ec8fd254992e5fa81d375e60849993a81db5e9e42a79344ebff7a40a0b45f"},
]

[package.dependencies]
langchain-core = ">=0.3,<0.4"
openai = ">=1.40.0,<2.0.0"
tiktoken = ">=0.7,<1"

saghul: Hum, weird we have a direct dependency on ^1.12. Perpahs we should drop it, since we don't directly use iot?

quitrk: probably a remnant from a previous implementation. removed

PR #103, opened 2024-10-03
feat: switch to using vllm

quitrk:
* when cuda is not available, we are still using llama.cpp as a fallback


pyproject.toml
```diff
@@ -22,16 +22,16 @@ pytest-mock = "3.12.0"
...
-prometheus-fastapi-instrumentator = "6.1.0"
+prometheus-client = "0.20.0"
+prometheus-fastapi-instrumentator = "7.0.0"
 pyjwt = {extras = ["crypto"], version = "^2.8.0"}
 python = "~3.11"
 redis = "5.0.1"
-torch = ">=2.0.0,<2.1.0"
-torchaudio = ">=2.0.1,<2.1.0"
+torch = "2.4.0"
+torchaudio = "2.4.0"
```
quitrk, line 34: @rpurdel i tested this with the github demo page and it seems to be working, anything else I could do to ensure I'm not breaking stuff?

skynet/modules/ttt/openai_api/app.py
```diff
@@ -23,16 +26,28 @@ def initialize():
...
-            --ctx-size {llama_n_ctx} \
-            --flash-attn \
-            --model {llama_path} \
-            --n-gpu-layers {llama_n_gpu_layers} \
-            --port {openai_api_server_port}'.split(),
-        shell=False,
-    )
+    if use_vllm:
+        openai_api_server_path = vllm_server_path
+        proc = subprocess.Popen(
```
quitrk, line 31: @saghul this allows us to restart just the api server when something times out

run.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/sh
-cd llama.cpp
-make llama-server
-cd ..
 
-export LLAMA_N_CTX=32000
+if nvcc --version
```
saghul, line 3: Doesn't this fail if you don't have the binary?

run.sh
```diff
@@ -1,7 +1,13 @@
...
+then
+  export CUDA_VISIBLE_DEVICES=0
```
saghul, line 5: Does 0 mean all? Or the first one, which is index 0?

skynet/env.py
```diff
@@ -87,7 +88,6 @@ def tobool(val: str | None):
 job_timeout = int(os.environ.get('JOB_TIMEOUT', 60 * 5))  # 5 minutes default
 
 # summaries
-summary_default_hint_type = os.environ.get('SUMMARY_DEFAULT_HINT_TYPE', 'text')
```
saghul, line 90: Why are we removing this? Is it related to the vllm change?

skynet/modules/ttt/openai_api/app.py
saghul: No strong opinion, we can always revisit after we gain more operational experience with vllm.

skynet/modules/ttt/openai_api/app.py
```diff
@@ -23,16 +26,28 @@ def initialize():
...
+            f'python -m {openai_api_server_path} \
+                --model {llama_path} \
+                --gpu_memory_utilization 0.95 \
+                --max-model-len {llama_n_ctx} \
+                --port {openai_api_server_port}'.split(),
```
saghul, line 36: Now that I see this, is it possible to limit the interface to localhost?

skynet/modules/ttt/summaries/processor.py
```diff
@@ -33,9 +33,11 @@ def initialize():
     global llm
 
     llm = ChatOpenAI(
+        model=llama_path,
         api_key='placeholder',  # use a placeholder value to bypass validation, and allow the custom base url to be used
         base_url=f'{openai_api_base_url}/v1',
         default_headers={"X-Skynet-UUID": app_uuid},
+        frequency_penalty=1,
```
saghul, line 40: What's this for?

skynet/modules/ttt/summaries/processor.py
```diff
@@ -33,9 +33,11 @@ def initialize():
     global llm
 
     llm = ChatOpenAI(
+        model=llama_path,
```
saghul, line 36: Do we need the full path? Or `os.path.basename(llama_path)` works?

skynet/modules/ttt/summaries/v1/models.py
```diff
@@ -14,7 +12,7 @@ class HintType(Enum):
 
 class DocumentPayload(BaseModel):
     text: str
-    hint: HintType = summary_default_hint_type
+    hint: HintType = HintType.TEXT
```
saghul, line 15: Ah I see. Good call. Arguably though, we focus more on conversations, so maybe the other default is better?

skynet/utils.py
```diff
@@ -46,3 +47,11 @@ async def create_webserver(app, port):
     )
     server = uvicorn.Server(server_config)
     await server.serve()
+
+
+def get_device() -> str:
```
saghul, line 52: These logs are going to be emitted every time the function is called. I'd cache the result.

saghul (commented):
Left some comments, mostly just questions, excellent stuff! Can you please also update the README with what AWQ model one should run when using vllm?

run.sh
quitrk: first one, this is only for local dev anyway

run.sh
quitrk: yeah, and it goes on the `else`

skynet/env.py
quitrk: not removed, but the default value type (string) does not match the declared pydantic type, which was an Enum, so I moved it there to use the local declaration of that Enum

skynet/modules/ttt/summaries/processor.py
quitrk: https://platform.openai.com/docs/advanced-usage/parameter-details

I used this to remove cases where the llm started hallucinating and repeating the same stuff over and over until it filled the context window

run.sh
saghul: Ah right.

skynet/modules/ttt/summaries/processor.py
saghul: Nice!

skynet/modules/ttt/summaries/processor.py
quitrk: what would that return when in Dockerfile we specify it as `LLAMA_PATH="/models/Llama-3.1-8B-Instruct-Q8_0.gguf"`?

skynet/modules/ttt/summaries/v1/models.py
quitrk: currently there's hardly and difference between them, if at all

skynet/modules/ttt/summaries/processor.py
saghul: `Llama-3.1-8B-Instruct-Q8_0.gguf` basename is the last part of a path.

saghul: Does it need the full path?

skynet/utils.py
quitrk: This is just a function I've moved from the streaming whisper module, from what I'm seeing it is only used at startup

skynet/modules/ttt/summaries/processor.py
quitrk: yup

skynet/modules/ttt/openai_api/app.py
quitrk: Not sure I understand this correctly but we could remove the [openai_api_base_url](https://github.com/jitsi/skynet/blob/master/skynet/env.py#L40) from being exposed as an env variable

quitrk:
> Left some comments, mostly just questions, excellent stuff! Can you please also update the README with what AWQ model one should run when using vllm?

We're not running AWQ as it is not optimised anyway, we're running raw, and still has some decent performance improvement over llama.cpp with quantized

skynet/utils.py
saghul: Ack!

skynet/modules/ttt/summaries/processor.py
saghul: The disregard my comment :-P

skynet/modules/ttt/openai_api/app.py
saghul: So, when you run this, does it bind to 0.0.0.0:PORT or to 127.0.0.1:PORT? The former would expose the service to the internet, if there is no port filtering, the latter won't.



> we could remove the [openai_api_base_url](https://github.com/jitsi/skynet/blob/master/skynet/env.py#L40) from being exposed as an env variable

I'd keep it because it makes testing other models very easy if you run ollama locally for example and put the URL in there...

saghul: In fact, perhaps we should think about that as the "development" option rather than running llama-server. That's a different discussion though.

saghul: Awesome!

skynet/modules/ttt/summaries/v1/models.py
saghul: Ok.

saghul (approved):

skynet/modules/ttt/openai_api/app.py
```diff
@@ -67,7 +69,11 @@ async def is_ready():
 def destroy():
     log.info('Killing OpenAI API subprocess...')
 
-    proc.kill()
+    if use_vllm:
+        subprocess.run(f'nvidia-smi --query-compute-apps=pid --format=csv,noheader | xargs -n1 kill -9', shell=True)
```
saghul, line 73: Don't we want to do this for any backend?

skynet/env.py
```diff
@@ -85,7 +85,7 @@ def tobool(val: str | None):
 
 
 # jobs
-job_timeout = int(os.environ.get('JOB_TIMEOUT', 60 * 5))  # 5 minutes default
+job_timeout = int(os.environ.get('JOB_TIMEOUT', 5))  # 5 minutes default
```
saghul, line 88: Oh this was in minutes?

run.sh
```diff
@@ -9,5 +9,5 @@ else
   cd ..
 fi
 
-export LLAMA_N_CTX=8192
+export LLAMA_N_CTX=17000
```
saghul, line 12: Is this intentional?

pyproject.toml
quitrk: tested on A10, seems to be working fine

PR #107, opened 2024-10-14
feat: add support for echoing requests

quitrk:
* useful for testing the impact of code changes between two different deployments with the same traffic data

### DIFF
--- a/skynet/modules/ttt/summaries/app.py
+++ b/skynet/modules/ttt/summaries/app.py
@@ -1,7 +1,8 @@
-import requests
+import random
 from fastapi import Request
 from fastapi_versionizer.versionizer import Versionizer
 
+from skynet import http_client
 from skynet.auth.openai import setup_credentials
 from skynet.env import echo_requests_base_url, echo_requests_percent, echo_requests_token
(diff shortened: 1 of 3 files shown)


skynet/modules/ttt/summaries/app.py
```diff
@@ -1,6 +1,9 @@
+import requests
```
saghul, line 1: We don't use this anywhere else do we? In fact we use aiohttp and a small client wrapper around it. Let's pick one. IIRC requests is blocking, so not a good fit here.

skynet/modules/ttt/summaries/app.py
```diff
@@ -16,6 +19,26 @@
...
 app.include_router(v1_router)
 
+post_requests_counter = 0
+
+if echo_requests_base_url:
+
+    @app.middleware("http")
+    async def echo_requests(request: Request, call_next):
+        if request.method == 'POST':
+            global post_requests_counter
```
saghul, line 29: Rather than a globakl counter, we can use a random number and roll the dice on each request.

PR #112, opened 2024-10-24
feat: expose openai api endpoints from vllm

quitrk:


skynet/modules/ttt/openai_api/app.py
```diff
@@ -9,37 +12,76 @@
...
+    from vllm.entrypoints.openai.api_server import (
+        build_async_engine_client,
+        init_app_state,
+        lifespan,
+        router,
+        TIMEOUT_KEEP_ALIVE,
+    )
+
+    async with build_async_engine_client(args) as engine_client:
+        app = create_app(lifespan=lifespan)
```
saghul, line 35: Looks like creating the app itself doesn't require any vllm async stuff, does it? If so we could reverse the logic, which is IMHO cleaner.

That is, create the app like we do others, then mount it in the main file, and pass it for further initialization, WDYT?

skynet/modules/ttt/openai_api/app.py
```diff
@@ -9,37 +12,76 @@
...
+        app.include_router(router, dependencies=dependencies, responses=responses)
+
+        model_config = await engine_client.get_model_config()
+        init_app_state(engine_client, model_config, app.state, args)
+
+        main_app.mount('/openai', app)
+
+        shutdown_task = await serve_http(
```
saghul, line 43: Is this necessary? Since the app is served by our uvicorn, what does this serve_http do? Why do we need to specify port and host and everything?

skynet/modules/ttt/openai_api/app.py
quitrk: wanted to centralize the checks for whether vllm or llama need to be loaded

skynet/env.py
```diff
@@ -1,12 +1,27 @@
...
+from functools import lru_cache
+
+import torch
 
 app_uuid = str(uuid.uuid4())
 
 is_mac = sys.platform == 'darwin'
 
 
+@lru_cache(maxsize=100)
```
saghul, line 13: Why not use a global variable which starts as None and is set only once?

skynet/env.py
```diff
@@ -1,12 +1,27 @@
...
+def get_device() -> str:
+    if torch.cuda.is_available():
+        print('CUDA device found.')
```
saghul, line 16: Use the logger?

skynet/env.py
```diff
@@ -1,12 +1,27 @@
...
+        return 'cuda'
+    print('No CUDA device found, defaulting to CPU.')
```
saghul, line 18: Ditto

skynet/env.py
```diff
@@ -36,9 +52,10 @@ def tobool(val: str | None):
 
 # openai api
 llama_cpp_server_path = os.environ.get('LLAMA_CPP_SERVER_PATH', './llama.cpp/llama-server')
-vllm_server_path = os.environ.get('VLLM_SERVER_PATH', 'vllm.entrypoints.openai.api_server')
-openai_api_server_port = int(os.environ.get('OPENAI_API_SERVER_PORT', 8003))
-openai_api_base_url = os.environ.get('OPENAI_API_BASE_URL', f'http://localhost:{openai_api_server_port}')
+openai_api_server_port = int(os.environ.get('OPENAI_API_SERVER_PORT', app_port if use_vllm else 8003))
```
saghul, line 55: Shall we get rid of this while we're here, and use the URL, defaulting to ollama's default port when not doing vllm?

skynet/env.py
quitrk: didnt want to create circular deps, is that a thing to worry about?

skynet/main.py
```diff
@@ -7,18 +7,23 @@
...
 
 if not modules:
-    log.warn('No modules enabled!')
+    log.warning('No modules enabled!')
     sys.exit(1)
 
 log.info(f'Enabled modules: {modules}')
 
+if device == 'cuda' or is_mac:
+    log.info('Using GPU')
```
saghul, line 23: I like how you solved this!

saghul (approved):
LGTM! We can get rid off the llama-cpp-server thing in a separate PR...

PR #115, opened 2024-10-30
fix: healthchecks failing due to missing internal id

quitrk:


run.sh
```diff
@@ -8,4 +8,4 @@ else
   export LLAMA_N_CTX=8182
 fi
 
-poetry run python -m uvicorn skynet.main:app --reload
+poetry run python -m skynet.main
```
saghul, line 11: Why the change?

saghul (approved):

run.sh
quitrk: to mirror what we are running the deployment, currently it was bypassing the normal startup flow from the `main` module

saghul: aha!

PR #116, opened 2024-10-31
fix: re-enable vLLM multiprocessing

quitrk:
* revert to launching vLLM as a separate process
* use vLLM router to expose OpenAI like APIs

### DIFF
--- a/skynet/env.py
+++ b/skynet/env.py
@@ -39,8 +39,9 @@ llama_n_ctx = int(os.environ.get('LLAMA_N_CTX', 128000))
 azure_openai_api_version = os.environ.get('AZURE_OPENAI_API_VERSION', '2024-02-01')
 
 # openai api
+openai_api_server_port = int(os.environ.get('OPENAI_API_SERVER_PORT', 8003))
 openai_api_base_url = os.environ.get(
-    'OPENAI_API_BASE_URL', f'http://localhost:{app_port}{"/openai"}' if use_vllm else "http://localhost:11434"
+    'OPENAI_API_BASE_URL', f'http://localhost:{openai_api_server_port}' if use_vllm else "http://localhost:11434"
 )
 
 # openai
--- a/skynet/env.py
+++ b/skynet/env.py
@@ -39,9 +39,9 @@ llama_n_ctx = int(os.environ.get('LLAMA_N_CTX', 128000))
 azure_openai_api_version = os.environ.get('AZURE_OPENAI_API_VERSION', '2024-02-01')
 
 # openai api
-openai_api_server_port = int(os.environ.get('OPENAI_API_SERVER_PORT', 8003))
+vllm_server_port = int(os.environ.get('VLLM_SERVER_PORT', 8003))
 openai_api_base_url = os.environ.get(
-    'OPENAI_API_BASE_URL', f'http://localhost:{openai_api_server_port}' if use_vllm else "http://localhost:11434"
+    'OPENAI_API_BASE_URL', f'http://localhost:{vllm_server_port}' if use_vllm else "http://localhost:11434"
 )
 
 # openai
--- a/skynet/http_client.py
+++ b/skynet/http_client.py
@@ -33,4 +33,17 @@ async def post(url, **kwargs):
         return await response.json()
 
 
-__all__ = ['get']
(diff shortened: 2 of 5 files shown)


skynet/http_client.py
```diff
@@ -33,4 +33,17 @@ async def post(url, **kwargs):
...
+    session = _get_session()
+
+    return await session.request(method, url, **kwargs)
+
+
+async def close():
+    session = _get_session()
+
+    if session is not None:
+        await _session.close()
```
saghul, line 46: shouldn't we set it to None after closing?

skynet/env.py
```diff
@@ -39,8 +39,9 @@ def tobool(val: str | None):
 azure_openai_api_version = os.environ.get('AZURE_OPENAI_API_VERSION', '2024-02-01')
 
 # openai api
+openai_api_server_port = int(os.environ.get('OPENAI_API_SERVER_PORT', 8003))
```
saghul, line 42: Let's call it `vllm_server_port` since it's not generic.

skynet/main.py
```diff
@@ -63,6 +62,8 @@ async def lifespan(main_app: FastAPI):
 
         await executor_shutdown()
 
+    await http_client.close()
```
saghul, line 65: Just for my education here, why do we need this now?

skynet/modules/ttt/openai_api/app.py
```diff
@@ -1,69 +1,85 @@
...
-                '--max-model-len',
-                str(llama_n_ctx),
-                '--port',
-                str(app_port),
-            ]
-        )
-
-        asyncio.create_task(run_vllm_server(args, app))
+    proc = subprocess.Popen(
+        f'python -m vllm.entrypoints.openai.api_server \
```
saghul, line 26: Rather than `python` let's use `sys.executable` it's more infallible.

skynet/modules/ttt/openai_api/app.py
```diff
@@ -1,69 +1,85 @@
...
+
+    if not bypass_auth:
+        try:
+            await bearer.__call__(request)
+        except HTTPException as e:
+            return JSONResponse(content=responses.get(e.status_code), status_code=e.status_code)
+
+    try:
+        url = f'{openai_api_base_url}{request.url.path.replace("/openai", "")}'
+        response = await http_client.request(request.method, url, headers=request.headers, data=await request.body())
```
saghul, line 76: Doesn't this break streaming since we are awaiting for the whole body?

skynet/modules/ttt/openai_api/app.py
```diff
@@ -1,69 +1,85 @@
...
+
+        return StreamingResponse(response.content, status_code=response.status, headers=response.headers)
+    except ClientConnectorError as e:
+        return JSONResponse(content=str(e), status_code=500)
+    except HTTPException as e:
```
saghul, line 81: Does it make sense to add a last resort `except Exception` at the end here?

saghul (commented):
Good work! I left some questions / comments.

skynet/http_client.py
quitrk: i guess it doesn't really matter since this happens before the app closes

skynet/main.py
quitrk: we don't really need it, it bugged me it was throwing an exception when i was closing the local instance

skynet/modules/ttt/openai_api/app.py
quitrk: is there a case where you wanna stream the request?

skynet/http_client.py
saghul: Right, I noticed that after seeing where it's used. Still, it IMHO makes sense to have its functionality be explicitly correct, not implicitly because of how we use it.

skynet/main.py
saghul: Great, makes sense to fix it!

skynet/modules/ttt/openai_api/app.py
saghul: 🤦  True, I guess no, so we're good!

saghul (approved):

PR #119, opened 2024-11-13
feat: add fallback folder when looking up public keys

quitrk:


skynet/auth/jwt.py
```diff
@@ -5,22 +5,44 @@
...
+        url = f'{asap_pub_keys_url}/{asap_pub_keys_fallback_folder}/{pub_key_remote_filename}'
+
+        log.info(f'Fetching public key {kid} from {url}')
+        key = await http_client.get(url, 'text')
+
+        if is_valid_key(key):
+            return key
 
-    return await http_client.get(url, 'text')
+    raise Exception()
```
saghul, line 45: Let's givit it a description perhaps? "Couldn't find key" or something.

skynet/auth/jwt.py
```diff
@@ -37,7 +59,7 @@ async def authorize(jwt_incoming: str) -> dict:
     try:
         public_key = await get_public_key(kid)
     except Exception:
-        raise HTTPException(status_code=401, detail=f'Failed to retrieve public key. {kid}')
+        raise HTTPException(status_code=401, detail=f'Failed to retrieve public key {kid}')
```
saghul, line 62: Ah I see. We could throw above with that description since that is where it originates, and here reuse the exception message.

saghul (approved):
LGTM with a suggestion.

saghul (approved):

PR #123, opened 2024-11-18
fix: remove public key validation

quitrk:


skynet/auth/jwt.py
```diff
@@ -17,32 +17,33 @@
...
-        log.info(f'Fetching public key {kid} from {url}')
-        key = await http_client.get(url, 'text')
+                log.info(f'Fetching public key {kid} from {url}')
+
+                response = await http_client.request('GET', url)
 
-        if is_valid_key(key):
-            return key
+                if response.status != 200:
+                    raise Exception()
```
saghul, line 40: Let's put the message here and then raise

PR #124, opened 2024-11-19
Rpurdel/upgrade vad

rpurdel:


skynet/modules/stt/streaming_whisper/cfg.py
```diff
@@ -11,12 +10,11 @@
     whisper_model_path,
 )
 from skynet.logs import get_logger
-from skynet.modules.stt.streaming_whisper.utils import vad_utils as vad
 
 log = get_logger(__name__)
 
 
-vad_model = vad.init_jit_model(f'{os.getcwd()}/skynet/modules/stt/streaming_whisper/models/vad/silero_vad.jit')
+vad_model = load_silero_vad(onnx=False)
```
saghul, line 17: Can you remind me why we don't want to use onnx here? That's the GPU acceleration, right?

saghul (approved):
❤️  that we are deleting all that code!

PR #130, opened 2024-12-06
Add autoscaling agent rest api

rpurdel:


skynet/agent.py
```diff
@@ -1,15 +1,23 @@
...
+from pydantic import BaseModel
+
 from skynet.env import whisper_max_connections
 
 from skynet.logs import get_logger
-from skynet.modules.monitoring import CONNECTIONS_METRIC
+from skynet.modules.monitoring import CONNECTIONS_METRIC, TRANSCRIBE_GRACEFUL_SHUTDOWN, TRANSCRIBE_STRESS_LEVEL_METRIC
 
 log = get_logger(__name__)
+app = FastAPI()
```
saghul, line 13: Since we are mounting a new app, I reckon we'd want to contain it in its own file perhaps, rather than together with this agent code?

skynet/agent.py
```diff
@@ -1,15 +1,23 @@
...
+
+STATE_FILE = '/tmp/streaming-whisper-connections-state'
```
saghul, line 15: Does anyone else read this file?

skynet/agent.py
```diff
@@ -1,15 +1,23 @@
...
+haproxy_state = 'ready'
+haproxy_check_state_task = None
 
 
-async def calc_percentage():
+async def get_haproxy_percentage():
```
saghul, line 20: Now that I look at this, there is nothing async inside this function, so why the async ?

skynet/agent.py
```diff
@@ -19,9 +27,9 @@ async def calc_percentage():
 
 
 async def handle_tcp_request(reader: asyncio.StreamReader, writer: asyncio.StreamWriter):
-    percentage = await calc_percentage()
-    status = 'up ready '
-    response = f'{status}{percentage}%\n'
+    global haproxy_state
+    percentage = await get_haproxy_percentage()
```
saghul, line 31: See above, looks like we can perhaps drop the await.

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+    desired_state: str
+
+
+class CurrentStateResponse(BaseModel):
+    current_state: str
+    active_connections: int
+    stress_level: float
+
+
+async def get_current_state():
```
saghul, line 67: This function blocks the asyncio loop because you are using standard library funcions, which block. You need to use aiofiles for asyncio friendly filesystem APIs.

More importantly, why do we need to read the file every time it's queried? What would cause its value to change? Is anyone else writing to that file?

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+    try:
+        with open(STATE_FILE, 'r') as f:
```
saghul, line 69: Rather than opening the file twice in case it's not there, you can open it for reading _and_ writing, and return the content if you find any, or write it if missing.

Not sure how often that operation happens though...

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+        with open(STATE_FILE, 'w') as f:
+            f.write('ready')
+        return 'ready'
+    except Exception as e:
+        log.error(f'Failed to read the state file: {e}')
+        return 'unknown'
+
+
+@app.post('/state')
+async def echo_requests(request: Request):
```
saghul, line 82: Weird function name? What does it have to do with echoing requests?

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+    payload = await request.json()
+    log.info(f'Received state change request: {payload}')
+    current_state = await get_current_state()
+
+    if haproxy_check_state_task is None:
+        loop = asyncio.get_running_loop()
+        haproxy_check_state_task = loop.create_task(update_haproxy_state())
+
+    if payload['state'] and payload['state'] in ['drain', 'maint', 'ready']:
+        with open(STATE_FILE, 'w') as f:
```
saghul, line 93: What happens if you fail to write the file?

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+            f.write(payload['state'].strip())
+        return StateResponse(request_status='success', current_state=current_state, desired_state=payload['state'])
+
+    log.error(f'Invalid state change request: {payload}')
+    return StateResponse(request_status='failed', current_state=current_state, desired_state=payload['state'])
+
+
+@app.get('/state')
+async def echo_requests(request: Request):
```
saghul, line 102: Ditto weird function name

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+    return CurrentStateResponse(
+        current_state=await get_current_state(),
```
saghul, line 104: See above, I think we can drop await

skynet/agent.py
```diff
@@ -41,3 +49,75 @@ async def create_tcpserver(port):
...
+# Updates the HAProxy state if it changes via the API
+# so that the load balancer can react to the changes
+async def update_haproxy_state():
+    global haproxy_state
+    while True:
+        new_state = await get_current_state()
+        if new_state != haproxy_state:
+            log.info(f'HAProxy state changed from {haproxy_state} to {new_state}')
+            haproxy_state = new_state
+        if new_state in ('drain', 'maint'):
```
saghul, line 119: Are we supposed to do anything in this state, such as reject new incoming requests?

skynet/agent.py
```diff
@@ -1,15 +1,23 @@
 import asyncio
```
saghul, line 1: Can we pl rename this file to haproxy_agent.py? The "agent" term is a bit overloaded in the AI space :-)

saghul (commented):
Some comments!

skynet/agent.py
rpurdel: They need to communicate with one another (that was initially the reason for the temp file as they were in separate modules). I dropped the temp file as they can share a global. If the autoscaler sets the app's state to **drain**, that state needs to be reflected in the HAProxy agent response. So the LB can stop sending new connections to this machine. No need to stop accepting connections from Skynet itself as HAProxy will handle it.

skynet/haproxy_agent.py
```diff
@@ -0,0 +1,97 @@
...
+
+async def handle_tcp_request(reader: asyncio.StreamReader, writer: asyncio.StreamWriter):
+    response = f'up {haproxy_state} {get_haproxy_lb_percentage()}%\n'
+    if writer is not None:
+        try:
+            writer.write(response.encode())
+        except Exception as ex:
+            log.warning('TCP exception during haproxy-agent check.', ex)
+        finally:
+            await writer.drain()
```
saghul, line 40: If the TCP connection is broken, won't this fail?

skynet/haproxy_agent.py
```diff
@@ -0,0 +1,97 @@
...
+from skynet.modules.monitoring import CONNECTIONS_METRIC, TRANSCRIBE_GRACEFUL_SHUTDOWN, TRANSCRIBE_STRESS_LEVEL_METRIC
+
+log = get_logger(__name__)
+
+# Bundles a REST API server used by the autoscaler with a TCP server that is queried by the HAProxy agent.
+# The REST API is used to change and query the state of the system. The HAProxy agent server is closely
+# bound to the REST API server as the latter will inform HAProxy if the system was set to drain mode.
+
+autoscaler_rest_app = FastAPI()
+TRANSCRIBE_GRACEFUL_SHUTDOWN.set(0)
```
saghul, line 19: Is it necwesssary to initialize it?

saghul (approved):
LGTM! Left 2 non-blocking questions, feel free to land it as-is.

skynet/haproxy_agent.py
rpurdel: nope

PR #134, opened 2025-01-14
Update faster-whisper to 1.1.1

rpurdel:


rpurdel:
mostly performance improvements

PR #135, opened 2025-01-20
ref: dettach logic for selecting inference processor from the jobs ru…

quitrk:
…nner

This will allow us to reuse this logic for the assistant usecase


saghul (approved):
Good one! 👍

PR #138, opened 2025-01-30
feat: implement RAG

quitrk:


skynet/modules/ttt/processor.py
```diff
@@ -96,28 +109,68 @@ def get_local_llm(**kwargs):
...
     )
 
 
+compressor = FlashrankRerank()
+
+
+async def assist(payload: DocumentPayload, customer_id: str | None = None, model: ChatOpenAI = None) -> str:
+    current_model = model or get_local_llm(max_completion_tokens=payload.max_completion_tokens)
+
+    if customer_id:
```
saghul, line 126: Shouldn't we give an error otherwise?

skynet/modules/ttt/processor.py
```diff
@@ -96,28 +109,68 @@ def get_local_llm(**kwargs):
...
+        store = await get_vector_store()
+        vector_store = await store.get(customer_id)
+
+        if vector_store:
```
saghul, line 130: Ditto

skynet/modules/ttt/processor.py
```diff
@@ -150,19 +203,29 @@ async def summarize(payload: DocumentPayload, job_type: JobType, model: BaseChat
...
+    payload: DocumentPayload, job_type: JobType, api_key: str, model_name=None, customer_id: str | None = None
+) -> str:
     llm = ChatOpenAI(
         api_key=api_key,
         max_completion_tokens=payload.max_completion_tokens,
         model_name=model_name,
         temperature=0,
     )
 
+    if job_type == JobType.ASSIST:
```
saghul, line 216: I feel like we could avoid doing the same logic if we'd make these take a function as a parameter, some `process_func`.

We can do that at a later time.

That would also mean that we can run on OCI to process too, which this PR doesn't account for, for example.

skynet/modules/ttt/rag/stores/faiss.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class FAISSVectorStore(SkynetVectorStore):
+    def __init__(self):
+        super().__init__()
+
+        if use_s3:
+            from skynet.modules.ttt.rag.stores.s3 import RagS3
+
+            log.info('Using S3 for vector store')
```
saghul, line 27: Can you reword this, since S3 is not the actual vector store but rather where we store the vestor store? (yeah that sounded weird in my head too)

skynet/modules/ttt/rag/stores/faiss.py
```diff
@@ -0,0 +1,82 @@
...
+            await self.s3.replicate(self.get_vector_store_path)
+
+    async def get(self, store_id):
+        try:
+            return FAISS.load_local(
+                self.get_vector_store_path(store_id), self.embedding, allow_dangerous_deserialization=True
+            )
+        except RuntimeError:
+            log.info(f'Vector store with id {store_id} not found')
+            return None
```
saghul, line 46: Whay is the caller supposed to do if they get None?

skynet/modules/ttt/rag/stores/faiss.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+        uuids = [str(uuid4()) for _ in range(len(documents))]
+        await vector_store.aadd_documents(documents=documents, ids=uuids)
+        vector_store.save_local(self.get_vector_store_path(store_id))
+        end = time.perf_counter_ns()
+        duration = round((end - start) / 1e9)
+
+        if use_s3:
+            self.s3.upload(self.get_vector_store_path(store_id))
```
saghul, line 66: Is this blocking? If not, does it need an await?

skynet/modules/ttt/rag/vector_store.py
```diff
@@ -0,0 +1,122 @@
...
+from skynet.modules.ttt.assistant.v1.models import RagConfig, RagPayload, RagStatus
+from skynet.modules.ttt.rag.constants import ERROR_RAG_KEY, RUNNING_RAG_KEY, STORED_RAG_KEY
+from skynet.modules.ttt.rag.web_crawler.main import crawl
+
+from ..persistence import db
+
+log = get_logger(__name__)
+
+
+class SkynetVectorStore(ABC):
```
saghul, line 20: LOok at you creating ABCs! ❤️

skynet/modules/ttt/rag/vector_store.py
```diff
@@ -0,0 +1,122 @@
...
+    embedding = HuggingFaceEmbeddings(
+        model_name=embeddings_model_path, model_kwargs={'device': 'cpu', 'trust_remote_code': True}
```
saghul, line 22: Why was trust_remote_code needed?

skynet/modules/ttt/rag/web_crawler/main.py
```diff
@@ -0,0 +1,72 @@
...
+async def crawl_url(url: str, max_depth) -> List[Document]:
+    """
+    Crawl the given URL and return the list of documents.
+    """
+
+    start = time.perf_counter_ns()
+    log.info(f"Starting to crawl {url}")
+
+    loader = SkynetRecursiveUrlLoader(
+        url=url, extractor=extractor, headers={'User-Agent': ua.random}, max_depth=max_depth, use_async=True, timeout=30
```
saghul, line 37: Is that `ua.random` a good idea?

skynet/modules/ttt/rag/web_crawler/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+async def crawl(payload: RagPayload) -> List[Document]:
+    """
+    Crawl the given URLs and return the concatanated list of documents.
+    """
+
+    documents = []
+    for url in payload.urls:
+        docs = await crawl_url(url, payload.max_depth)
```
saghul, line 66: Here you are crawling them serialized, is that really what you want? You could run them concurrently and wait for all the futures to end and gather their results.

skynet/modules/ttt/rag/web_crawler/recursive_url_loader.py
```diff
@@ -0,0 +1,112 @@
...
+        url: str,
+        visited: Set[str],
+        *,
+        session: Optional[aiohttp.ClientSession] = None,
+        depth: int = 0,
+    ) -> List[Document]:
+        if depth >= self.max_depth:
+            return []
+
+        # Disable SSL verification because websites may have invalid SSL certificates,
```
saghul, line 32: Smart!

skynet/modules/ttt/s3.py
```diff
@@ -0,0 +1,66 @@
...
+                's3',
+                aws_access_key_id=skynet_s3_access_key,
+                aws_secret_access_key=skynet_s3_secret_key,
+                endpoint_url=skynet_s3_endpoint,
+                region_name=skynet_s3_region,
+            )
+        except Exception as e:
+            log.error(f'Failed to connect to S3: {e}')
+
+    def download_file(self, filename):
```
saghul, line 28: This looks blocking which is no bueno. I'd recommend having a dedicated executor or using the current loop's to run the fs ops and turn them into async.

saghul: This should apply for all the functions here.

saghul (commented):
Impressive work! 👏  Left some comments, PTAL!

skynet/env.py
```diff
@@ -128,3 +131,17 @@ def tobool(val: str | None):
 oci_compartment_id = os.environ.get('OCI_COMPARTMENT_ID')
 oci_auth_type = os.environ.get('OCI_AUTH_TYPE', 'API_KEY')
 oci_config_profile = os.environ.get('OCI_CONFIG_PROFILE', 'DEFAULT')
+
+# rag
+vector_store_path = os.environ.get('VECTOR_STORE_PATH', '_vector_store_')
+supported_vector_store_types = {'faiss'}
```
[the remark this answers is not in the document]
quitrk: the plan is to get rid of s3 and use something like else that allows us to do the search directly rather than replicating the whole db locally, thus i added this structure, so we can switch between modes when running locally

skynet/main.py
```diff
@@ -37,22 +39,26 @@ async def lifespan(main_app: FastAPI):
...
+
+        main_app.mount('/assistant', rag_app)
+        await assistant_startup()
+
     if 'summaries:dispatcher' in modules:
+        from skynet.modules.ttt.openai_api.app import app as openai_api_app
         from skynet.modules.ttt.summaries.app import app as summaries_app, app_startup as summaries_startup
 
         main_app.mount('/summaries', summaries_app)
+        main_app.mount('/openai', openai_api_app)
```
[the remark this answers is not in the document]
quitrk: let's revisit this one

skynet/modules/ttt/processor.py
```diff
@@ -96,28 +109,68 @@ def get_local_llm(**kwargs):
...
+
+    if customer_id:
+        store = await get_vector_store()
+        vector_store = await store.get(customer_id)
+
+        if vector_store:
+            base_retriever = vector_store.as_retriever(search_kwargs={'k': 3})
+            retriever = ContextualCompressionRetriever(base_compressor=compressor, base_retriever=base_retriever)
+
+            prompt_template = '''
```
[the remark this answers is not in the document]
quitrk: you mean for summaries and action items? maybe, but we should probably do it separately

skynet/modules/ttt/processor.py
quitrk: i'll remove the customer_id validation since it's redundant, we'll always have one on this branch of code, however there's still some missing implementation when there's no vector store.

skynet/modules/ttt/processor.py
quitrk: see above

skynet/modules/ttt/rag/stores/faiss.py
quitrk: rewording to `Using S3 for vector store persistence`

skynet/modules/ttt/processor.py
saghul: Got it. Let's give an error for now then?

skynet/modules/ttt/rag/stores/faiss.py
quitrk: that would be the `else` case I was talking about for `if vector_store:` in `processor`

skynet/modules/ttt/rag/vector_store.py
quitrk: "The trust_remote_code=True flag is used for models whose code lives on the hub rather than natively in the Transformers library."

skynet/modules/ttt/rag/web_crawler/recursive_url_loader.py
quitrk: can't take credit for this, I just monkey patched the original langchain method to add pdf parsing support, I'd gladly take another approach...

skynet/modules/ttt/rag/web_crawler/main.py
quitrk: I benchmarked both, and for some reason the asyncio.gather takes a bit longer, so I'm not sure it's worth the risk of potentially getting blocked due to some requests quota

skynet/env.py
```diff
@@ -31,23 +31,26 @@ def tobool(val: str | None):
...
+embeddings_model_path = os.environ.get('EMBEDDINGS_MODEL_PATH', 'nomic-embed-text')
+embeddings_model_n_ctx = int(os.environ.get('EMBEDDINGS_MODEL_N_CTX', 8192))
+
 # azure openai api
 # latest ga version https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation#latest-ga-api-release
 azure_openai_api_version = os.environ.get('AZURE_OPENAI_API_VERSION', '2024-02-01')
 
 # openai api
-vllm_server_port = int(os.environ.get('VLLM_SERVER_PORT', 8003))
+openai_api_port = int(os.environ.get('OPENAI_API_PORT', 8003))
```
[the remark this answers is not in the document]
quitrk: probably not

skynet/modules/ttt/rag/web_crawler/main.py
saghul: Fair point, let's keep it as is.

skynet/modules/ttt/processor.py
quitrk: i implemented the fallback, now we just provide an empty string to `Context` when no vector store

saghul (approved):
🔥

PR #141, opened 2025-02-07
feat: add ability to use OCI hosted models

saghul:


skynet/env.py
```diff
@@ -101,7 +101,7 @@ def tobool(val: str | None):
 
 # summaries
 summary_minimum_payload_length = int(os.environ.get('SUMMARY_MINIMUM_PAYLOAD_LENGTH', 100))
-enable_batching = tobool(os.environ.get('ENABLE_BATCHING', 'true'))
+enable_batching = tobool(os.environ.get('ENABLE_BATCHING', 'true' if use_vllm else 'false'))
```
quitrk, line 104: maybe `str(use_vllm)`?

skynet/modules/ttt/openai_api/app.py
```diff
@@ -40,6 +40,9 @@ def initialize():
 
 
 async def is_ready():
+    if llama_path.startswith('oci://'):
```
quitrk, line 43: we could extract this into a `use_oci` since we reuse this logic in the processor

skynet/env.py
quitrk: i assume it does not have anything to do with oci but rather with local inference with ollama?

saghul: Yeah it's for ollama. Good point on the simplification.

saghul:
@quitrk Updated, PTAL. I dropped the batching change since I doubt it has an impact in local testing...

skynet/modules/ttt/processor.py
```diff
@@ -55,7 +69,29 @@ def get_job_processor(customer_id: str) -> Processors:
...
 def get_local_llm(**kwargs):
+    # OCI hosted llama
+    if use_oci:
+        global oci_llm
+
+        if oci_llm is None:
+            oci_llm = ChatOCIGenAI(
+                model_id=oci_model_id,
+                service_endpoint=oci_service_endpoint,
+                compartment_id=oci_compartment_id,
```
saghul, line 86: For some reason this is necessary, the OCI python package complains otherwise.

quitrk (approved):

PR #146, opened 2025-02-13
feat: patch OCI library to be async

saghul:
The OCI client uses Requests under the hood, which is blocking. Patch it by running it in a thread pool, with a max size of 10, which matches the connection pool size.


saghul:
Closing, let's revisit at a later stage.

quitrk:
for future ref https://docs.python.org/dev/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

PR #147, opened 2025-02-13
Whisper check silence in buffer

rpurdel:
Moves the silence check to the buffer instead of the individual chunk. Which was imprecise due to its small size.

### DIFF
--- a/skynet/modules/stt/streaming_whisper/utils/utils.py
+++ b/skynet/modules/stt/streaming_whisper/utils/utils.py
@@ -196,7 +196,7 @@ def convert_seconds_to_bytes(cut_mark: float) -> int:
     return int(cut_mark / cfg.one_byte_s)
 
 
-def is_silent(audio: bytes) -> Tuple[bool, iter]:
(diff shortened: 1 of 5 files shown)


skynet/modules/stt/streaming_whisper/utils/utils.py
```diff
@@ -196,7 +196,7 @@ def convert_seconds_to_bytes(cut_mark: float) -> int:
     return int(cut_mark / cfg.one_byte_s)
 
 
-def is_silent(audio: bytes) -> Tuple[bool, iter]:
+async def is_silent(audio: bytes) -> Tuple[bool, iter]:
```
saghul, line 199: I see no await in there, what's async about it?

saghul (approved):
LGTM with a question.

skynet/modules/stt/streaming_whisper/utils/utils.py
rpurdel: leftover

PR #149, opened 2025-02-17
fix: only restart server if the local processor times out

saghul:
If upstream OpenAI or OCI timeout there is no need to restart Skynet.

### DIFF
--- a/skynet/modules/ttt/summaries/jobs.py
+++ b/skynet/modules/ttt/summaries/jobs.py
@@ -38,7 +38,8 @@ def restart():
 
     OPENAI_API_RESTART_COUNTER.inc()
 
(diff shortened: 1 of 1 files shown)


skynet/modules/ttt/summaries/jobs.py
```diff
@@ -38,7 +38,8 @@ def restart():
 
     OPENAI_API_RESTART_COUNTER.inc()
 
-    # rely on the supervisor to restart the process
+    # Rely on the supervisor to restart the process.
+    # TODO: consider just restring the vllm subprocess.
```
quitrk, line 42: typo in this comment

quitrk (approved):
good catch

PR #150, opened 2025-02-17
feat: use redis pubsub to sync s3 bucket changes across nodes

quitrk:


skynet/modules/ttt/rag/stores/s3.py
```diff
@@ -21,6 +23,16 @@ async def files_aiter():
 class RagS3:
     def __init__(self):
         self.s3 = S3()
+        self.listen_task = None
+
+    def __del__(self):
```
saghul, line 28: In general is not great to rely on `__del__` because it can be non-determinisstic when it runs, as it relies on GC.

An explicit `close()` or `dispose()` method would be better. Do you think we could do that?

saghul (approved):
LGTM with a suggestion. ❤️  the simplicity!

skynet/modules/ttt/rag/stores/s3.py
quitrk: haha, that's how I initially implemented it but then I thought of the destructor, will. revert the the OG

PR #156, opened 2025-02-19
feat: support configuring assistant system message

quitrk:


skynet/modules/ttt/assistant/constants.py
```diff
@@ -0,0 +1,6 @@
+assistant_default_system_message = "You are an AI assistant. You provide helpful suggestions"
+assistant_rag_question_extractor = """
+    Based on the provided text, formulate a proper question for using with RAG.
```
saghul, line 3: Do we want to put RAG in there or something more generic like "provided documents" ?

skynet/modules/ttt/assistant/v1/models.py
```diff
@@ -44,13 +52,18 @@ class RagConfig(RagPayload):
...
+
     model_config = {
         'json_schema_extra': {
             'examples': [
                 {
                     'text': 'User provided context here (will be appended to the RAG one)',
                     'prompt': 'User prompt here',
                     'max_completion_tokens': None,
+                    'hint': 'conversation',
+                    'use_only_rag_data': False,  # If True a vector store is available, only the RAG data will be used for assistance
```
saghul, line 66:
```suggestion
                    'use_only_rag_data': False,  # If True and a vector store is available, only the RAG data will be used for assistance
```

skynet/modules/ttt/utils.py
```diff
@@ -0,0 +1,25 @@
+from skynet.modules.ttt.assistant.constants import assistant_default_system_message, assistant_limit_data_to_rag
```
saghul, line 1: Maybe call this file rag_utils?

saghul (approved):
LGTM with some minor suggestions.

skynet/modules/ttt/assistant/constants.py
quitrk: worked fine from my tests, let's stick to this for now imo

skynet/modules/ttt/utils.py
quitrk: oh, i mistakenly put it directly in ttt/, wanted to place it in ttt/assistant

skynet/modules/ttt/assistant/constants.py
saghul: Sure!

saghul (approved):

PR #158, opened 2025-02-20
Refactor processors

saghul:


skynet/modules/ttt/processor.py
```diff
@@ -191,13 +118,14 @@ async def summarize(payload: DocumentPayload, job_type: JobType, model: BaseChat
     )
 
     # this is a rough estimate of the number of tokens in the input text, since llama models will have a different tokenization scheme
-    num_tokens = current_model.get_num_tokens(text)
+    num_tokens = model.get_num_tokens(text)
 
     # allow some buffer for the model to generate the output
+    # TODO: adjust this to the actual model's context window
```
saghul, line 124: We will need to keep a mapping between models and context sizes. Right now we can make it simple since gpt4o is128k just like llama 3.1 so the only one with a slightly small window is the local 8b where we trim it.

We can do that in a follow-up PR, I just left this here so we don't forget :-)

skynet/modules/ttt/oci/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    oci_compartment_id,
+    oci_config_profile,
+    oci_model_id,
+    oci_service_endpoint,
+    use_oci,
+)
+
+
+async def initialize():
+    if not use_oci:
```
quitrk, line 15: i think this should be initialized based on whether credentials for oci are present or not, since the `use_oci` flag only covers the cpu deployment usecase

quitrk: which leads me to believe that the flag does not represent correctly what it does

saghul: You mean to check if all options are given? Yeah I can do that!

saghul: Ah wait no. `use_oci` means skynet has been configured to use OCI as the LLM provider. But it can do that in addition to running vllm locally.

Will fix.

saghul: Pushed a fix!

skynet/modules/ttt/llm_selector.py
```diff
@@ -0,0 +1,107 @@
...
+
+            return AzureChatOpenAI(
+                api_key=options.get('secret'),
+                api_version=azure_openai_api_version,
+                azure_endpoint=metadata.get('endpoint'),
+                azure_deployment=metadata.get('deploymentName'),
+                max_completion_tokens=max_completion_tokens,
+                temperature=0,
+            )
+        elif processor == Processors.OCI or use_oci:
```
quitrk, line 73: shouldn't this become `and oci_available`? or have that check in the `get_job_processor` logic

saghul: yeah it makes sense!

saghul: done!

skynet/modules/ttt/processor_test.py
```diff
@@ -88,20 +92,52 @@ async def test_process_with_azure_open_ai(self, process_fixture):
...
+
+        process_fixture.patch(
+            'skynet.modules.ttt.llm_selector.get_credentials',
+            return_value={'type': 'OCI'},
+        )
+        process_fixture.patch('skynet.modules.ttt.llm_selector.oci_available', True)
+
+        job = Job(
+            payload=DocumentPayload(
+                text="Andrew: Hello. Beatrix: Honey? It’s me . . . Andrew: Where are you? Beatrix: At the station. I missed my train."
```
quitrk, line 116: queue this up https://www.youtube.com/watch?v=wMOkm57vu0k

quitrk (approved):

PR #162, opened 2025-02-20
feat(ci) add CI for testing Docker builds

saghul:


saghul:
My plan is to extend the release CI after to push a CPU image to DockerHub too, if we that's useful.

saghul:
```
#17 104.7 qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

Sigh. Now that Linux ARM runners are in beta I'll give it a try to do native builds.

PR #169, opened 2025-02-26
feat: implement slim openai api fallback

quitrk:


skynet/modules/ttt/openai_api/app.py
```diff
@@ -20,6 +21,7 @@
 
 def initialize():
     if not use_vllm:
+        app.include_router(slim_router)
```
saghul, line 24: do we also want to do it if there is a local llama? I'd only add it if `use_oci` is true

skynet/modules/ttt/openai_api/app.py
```diff
@@ -72,29 +74,31 @@ async def is_ready():
...
-        return JSONResponse(content=e.detail, status_code=e.status_code)
-    except Exception as e:
-        return JSONResponse(content=str(e), status_code=500)
-
+            return JSONResponse(content=e.detail, status_code=e.status_code)
+        except Exception as e:
+            return JSONResponse(content=str(e), status_code=500)
 
-__all__ = ['app', 'initialize', 'is_ready']
+    __all__ = ['app', 'initialize', 'is_ready']
```
saghul, line 104: Weird indent

skynet/modules/ttt/openai_api/slim_router.py
```diff
@@ -0,0 +1,38 @@
...
+    choices: List[ChatCompletionResponseChoice]
+
+
+class ChatCompletionRequest(BaseModel):
+    max_completion_tokens: Optional[int] = None
+    messages: List[ChatCompletionMessageParam]
+
+
+@router.post('/v1/chat/completions')
+async def create_chat_completion(chat_request: ChatCompletionRequest, request: Request):
```
saghul, line 33: Does this work with streaming requests?

quitrk: no, and i don't think our `ainvoke` does either since langchain exposes an `astream` method, i'm planning that for future times

quitrk: we only have 1 consumer and I already studied how they're using it, thus the small factor for now

saghul (approved):

PR #175, opened 2025-03-03
feat: implement support for custom locales

quitrk:


skynet/modules/ttt/processor.py
```diff
@@ -153,6 +155,7 @@ async def process_text(model: BaseChatModel, payload: DocumentPayload) -> str:
 
     prompt = ChatPromptTemplate(
         [
+            ('system', set_response_language(payload.preferred_locale)),
```
saghul, line 158: I would leave it out from process_text, the user should take care of any prompting.

skynet/modules/ttt/summaries/prompts/common.py
```diff
@@ -0,0 +1,12 @@
+from skynet.constants import Locale
+
+
+stick_to_main_language = "Respond using the main language of the provided text."
+response_prefix = "ProcessingResponse"
```
saghul, line 5: `SkynetResponse` so it never attempts to translate it?

saghul (approved):
LGTM with a couple of suggestions. Q: how would a user know if a language is supported? Pycdantic will fail the payload validation, right?

quitrk:
> LGTM with a couple of suggestions. Q: how would a user know if a language is supported? Pycdantic will fail the payload validation, right?

Well, there's a swagger I suppose. Pydantic will fail if the language is not present in the enum.

skynet/modules/ttt/processor.py
saghul: Nah, they get full prompt control, so the user is on the driver seat...

skynet/modules/ttt/summaries/prompts/common.py
saghul: 🔥

saghul:
> Pydantic will fail if the language is not present in the enum.

That's good enough!

PR #179, opened 2025-03-06
ci: add e2e tests to ci workflow

quitrk:


e2e/summaries.py
```diff
@@ -56,7 +51,7 @@ async def run():
     job = await create_summary()
     job_id = job.get('id')
 
-    await sleep_progress(5, 'Waiting for summary to be processed')
+    await sleep_progress(7, 'Waiting for summary to be processed')
```
saghul, line 54: can't we poll the endpoint with a timeout instead?

quitrk: hah, coincidentally i actually changed that to exactly this what you suggested

saghul (approved):

PR #180, opened 2025-03-07
ref: update process method signature to accept a Job object

quitrk:


saghul (approved):
Much better indeed!

quitrk:
> Much better indeed!

there's a hidden agenda behind this :)

PR #181, opened 2025-03-07
feat: fallback to local processing if OCI fails

quitrk:


skynet/modules/ttt/processor.py
```diff
@@ -174,15 +174,26 @@ async def process(job: Job) -> str:
...
+        elif job_type == JobType.PROCESS_TEXT:
+            result = await process_text(llm, payload)
+        else:
+            raise ValueError(f'Invalid job type {job_type}')
+    except Exception as e:
+        log.warning(f"Job {job.id} failed: {e}")
+
+        processor = LLMSelector.get_job_processor(customer_id, job.id)
+
+        if processor == Processors.OCI and not use_oci:
```
saghul, line 193: Should we make this generic? As in, if OPENAI fails we should also fallback, shouldn't we?

quitrk: thought we're not responsible for those 🤔

saghul (approved):

PR #188, opened 2025-03-12
feat: implement file uploads for rag

quitrk:


skynet/modules/ttt/rag/vector_store.py
```diff
@@ -105,26 +117,39 @@ async def get_config(self, store_id: str) -> Optional[RagConfig]:
...
-        Crawl the given URL and create a vector store with the generated embeddings.
+        Extract text from the payload and create a vector store with the generated embeddings.
         """
 
+        documents = []
+        error = None
+
         try:
-            documents = await crawl(payload)
+            from skynet.modules.ttt.rag.text_extractor.main import extract
```
saghul, line 129: Any reason to import these here rater than at the top?

saghul (approved):

skynet/modules/ttt/rag/vector_store.py
quitrk: not anymore

PR #194, opened 2025-03-19
feat: allow uploading zip files for RAG

quitrk:


saghul (approved):
Do we want to support recursion? That is, zip files inside zip files?

quitrk:
> Do we want to support recursion? That is, zip files inside zip files?

idk...if you want we could implement it

saghul:
> > Do we want to support recursion? That is, zip files inside zip files?
> 
> 
> 
> idk...if you want we could implement it

Let's just wait until we need it.

PR #195, opened 2025-03-25
feat: tweak assistant implementation

quitrk:
* retrieve top_k based on AssistantType


pyproject.toml
```diff
@@ -39,7 +39,6 @@ fake-useragent = "^2.0.3"
 fastapi = "0.115.11"
 fastapi-versionizer = "4.0.1"
 faster-whisper = "1.1.1"
-flashrank = "^0.2.10"
```
saghul, line 42: Can you elaborate on why removing this is A Good Thing?

skynet/env.py
```diff
@@ -140,7 +142,8 @@ def tobool(val: str | None):
 supported_vector_store_types = {'faiss'}
 vector_store_type = supported_vector_store_types.intersection({os.environ.get('VECTOR_STORE_TYPE', 'faiss').lower()})
 vector_store_type = vector_store_type.pop() if vector_store_type else None
-vector_store_top_k = int(os.environ.get('VECTOR_STORE_TOP_K', 5))
+assistant_search_top_k = int(os.environ.get('ASSISTANT_SEARCH_TOP_K', 3))
```
saghul, line 145: IMHO these belong as query parameters rather than have them as global settings.

skynet/modules/ttt/rag/stores/faiss.py
```diff
@@ -62,14 +63,20 @@ async def create(self, store_id, documents):
         index = faiss.IndexFlatL2(len(self.embedding.embed_query(store_id)))
         vector_store = FAISS(
             embedding_function=self.embedding,
-            distance_strategy=DistanceStrategy.EUCLIDEAN_DISTANCE,
+            distance_strategy=DistanceStrategy.COSINE,
             index=index,
             docstore=InMemoryDocstore(),
             index_to_docstore_id={},
+            normalize_L2=True,
```
saghul, line 70: What does this do?

skynet/modules/ttt/rag/stores/faiss.py
```diff
@@ -62,14 +63,20 @@ async def create(self, store_id, documents):
         index = faiss.IndexFlatL2(len(self.embedding.embed_query(store_id)))
         vector_store = FAISS(
             embedding_function=self.embedding,
-            distance_strategy=DistanceStrategy.EUCLIDEAN_DISTANCE,
+            distance_strategy=DistanceStrategy.COSINE,
```
saghul, line 66: Can we leave a comment at the top of this function about what we are doing and why?

skynet/modules/ttt/rag/vector_store.py
```diff
@@ -40,9 +40,7 @@ def bypass_ingestion(config: RagConfig, new_config: RagConfig):
 
 
 class SkynetVectorStore(ABC):
```
saghul, line 42: Why is that no longer needed?

pyproject.toml
quitrk: I'm still debating this myself. The idea is that better models already have good reranking capabilities, however, the impact of having this is so low that I don't think its worth removing it.

skynet/modules/ttt/rag/stores/faiss.py
quitrk: cosine is the recommended strategy for text search, however, we need to eliminate from its calculation the effect that the length of the vectors have on the result, because for similarity we only need the direction. This is where normalisation comes in

skynet/modules/ttt/rag/vector_store.py
quitrk: cause the model we're gonna switch to is backed by the standard sentence transformers

saghul: Then can we update the env var's default value, since it won't work otherwise?

PR #200, opened 2025-04-07
Skip OpenAI API health checking if using OpenAI or Azure API

vgerasymenko:
For now if using OpenAI or Azure API with credentials specified in _credentials.yaml_ passed via _SKYNET_CREDENTIALS_PATH_ env variable - _is_ready()_ will always return False and _monitor_candidate_jobs()_ in _modules\ttt\summaries\jobs.py_ will freeze forever in next cycle:

```
    while not await is_openai_api_ready():
        await asyncio.sleep(TIME_BETWEEN_JOBS_CHECK)
```

I've added condition which will return True in such case. This could be non-optimal (or incorrect) solution because using OpenAI/Azure API does not necessary means what local LLMs won't be used. May be _(device == 'cpu')_ condition should be used instead, which more likely means that local LLMs are not supposed to be used.


quitrk:
I'm not sure I understand the use case. Is this some scenario where you only forward to azure / openai? There's no check that can verify that since the whole idea is that different jobs can be processed by different providers.

vgerasymenko:
Yes, I plan to use only OpenAI provider for now. It does not work this way, since summary jobs will freeze forever on health checks. At least OpenAI seems to support some API about it's health, but I don't think it should be used.
I think there should be some additional parameter to indicate, that only AI service providers like OpenAI are planned to be used - so no health checks should be made. I've done that via _openai_credentials_file_ parameter checking which is not correct in general, as you mentioned.

quitrk:
*disable_llm_health_check* sounds alright to me

saghul (approved):

quitrk:
@vgerasymenko I've messed up something, could you please resolve the conflicts and do a rebase?

PR #201, opened 2025-04-14
feat: use oci as default when available

quitrk:


requirements-vllm.txt
```diff
@@ -1,205 +1,205 @@
-aioboto3==13.3.0 ; python_version >= "3.11" and python_version < "3.12"
```
saghul, line 1: Can you pl drop the dependency updates? Or make them in a separate PR if intentional.

saghul (approved):

PR #213, opened 2025-06-12
fix: proper update of the stress metric when both vox and streaming w…

rpurdel:
…hisper are used


rpurdel:
> Can you drop the dependency updates? They seem unrelated to the change.

done

PR #217, opened 2025-07-04
feat: add metric for map-reduce chunking operations

quitrk:
## Summary
- Added `MAP_REDUCE_CHUNKING_COUNTER` metric to track when long inputs trigger map-reduce chunking in the summarization pipeline
- The metric includes `job_type` label to distinguish between summary, action items, and table of contents operations
- This helps monitor performance and resource usage patterns for large document processing

## Changes
- Added new Prometheus counter metric in `skynet/modules/monitoring.py`
- Integrated metric tracking in `skynet/modules/ttt/processor.py` when chunking is triggered
- Metric increments when input text exceeds token threshold and requires map-reduce processing

## Test plan
- [ ] Deploy to staging environment
- [ ] Process large documents that trigger chunking
- [ ] Verify metric appears in Prometheus metrics endpoint
- [ ] Confirm metric labels are properly set

🤖 Generated with Claude Code

### DIFF
--- a/skynet/modules/monitoring.py
+++ b/skynet/modules/monitoring.py
@@ -105,6 +105,14 @@ OPENAI_API_RESTART_COUNTER = Counter(
     subsystem=PROMETHEUS_SUMMARIES_SUBSYSTEM,
 )
 
+MAP_REDUCE_CHUNKING_COUNTER = Counter(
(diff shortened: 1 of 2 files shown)


saghul (approved):
Maybe drop the prompt changes since the other PR does it?

PR #220, opened 2025-07-18
fix: path traversal when uploading files

quitrk:


skynet/modules/ttt/rag/utils.py
```diff
@@ -23,6 +25,34 @@ def split_documents(initial_documents: list[Document]) -> list[Document]:
     return splits
 
 
+def sanitize_filename(filename: str) -> str:
```
saghul, line 28: You should be able to just do os.path.abspath(os.path.realpath(the_path)) and then check if it's inside the folder you want, if not discard it, no?

quitrk: this won't prevent an attacker from accessing files from your system.

saghul: Not sure I get the vector. If we first get the full path then we can check if that path is allowed no?

quitrk: let's chat on Monday

quitrk: got some time to return to this, you were right

PR #227, opened 2025-09-03
feat(whisper): Track connections in a list, not keyed by meeting_id

aaronkvanmeerten:
This PR moves connection tracking to the lifecycle of a websocket instead of using the incoming session_id or meeting_id.  This avoids the problem of two connections with the same ID causing one to become unusable.


aaronkvanmeerten (commented):
This improves handling when multiple connections come in with the same session ID.  I've re-implemented so that nothing need search on the connections.

aaronkvanmeerten:
I've re-run the actions and all builds are OK.

PR #228, opened 2025-09-04
fix: add oci blackout when rate limited

quitrk:


skynet/modules/ttt/processor.py
```diff
@@ -201,6 +239,13 @@ async def process(job: Job) -> str:
     job_type = job.type
     customer_id = job.metadata.customer_id
 
+    # Check if OCI is currently in blackout period
+    processor = LLMSelector.get_job_processor(customer_id, job.id)
+    if processor == Processors.OCI and is_oci_blackout_active():
```
saghul, line 244: Shouldn't this logic be in the selector?

PR #230, opened 2025-09-08
feat: add customer config endpoints

horymury:
- added custom summary_prompt as a first prop of the config. It is used in case there is no prompt on the summary request itself


skynet/main.py
```diff
@@ -46,6 +46,15 @@ async def lifespan(main_app: FastAPI):
         main_app.mount('/assistant', rag_app)
         await assistant_startup()
 
+    if 'customerconfigs' in modules:
```
quitrk, line 49: not a fan of the combined wording, i would prefer the mounting path to be `customer-config`

skynet/modules/ttt/processor.py
```diff
@@ -157,11 +157,22 @@ async def assist(model: BaseChatModel, payload: AssistantDocumentPayload, custom
     return await rag_chain.ainvoke(input={'question': question})
 
 
-async def summarize(model: BaseChatModel, payload: DocumentPayload, job_type: JobType) -> str:
+async def summarize(model: BaseChatModel, payload: DocumentPayload, job_type: JobType, customer_id: str) -> str:
     chain = None
     text = payload.text
 
-    system_message = payload.prompt or hint_type_to_prompt[job_type][payload.hint]
+    # Fallback priority: payload.prompt -> customer's summary_prompt -> hint_type_to_prompt[job_type][payload.hint]
```
quitrk, line 164: should add some tests for these, have. claude generate them, we already got some for this part

